PR #7, opened 2018-08-15
RasterSourceRDD

jbouffard:
This PR adds the `RasterSourceRDD` to the `vlm` package. It should be noted that this feature is still a work in progress, and as such there may be issues when trying create an `RDD`.


vlm/src/main/scala/RasterSourceRDD.scala
```diff
@@ -0,0 +1,105 @@
...
+  )(implicit sc: SparkContext): MultibandTileLayerRDD[SpatialKey] = {
+    val mapTransform = layout.mapTransform
+    val extent = mapTransform.extent
+    val combinedExtents = sources.map { _.extent }.reduce { _ combine _ }
+
+    val cellType = {
+      val cellTypes = sources.map { source => Set(source.cellType) }.reduce { _ ++ _ }
+
+      if (cellTypes.size > 1)
+        throw new Exception(s"All RasterSources must have the same CellType, but multiple ones were found: $cellTypes")
```
echeipesh: Should use `require` here, its more idiomatic and will give a more clear exception.

vlm/src/main/scala/RasterSourceRDD.scala
```diff
@@ -0,0 +1,105 @@
...
+    sources: Seq[RasterSource],
+    layout: LayoutDefinition,
+    partitionBytes: Long = PARTITION_BYTES
+  )(implicit sc: SparkContext): MultibandTileLayerRDD[SpatialKey] = {
+    val mapTransform = layout.mapTransform
+    val extent = mapTransform.extent
+    val combinedExtents = sources.map { _.extent }.reduce { _ combine _ }
+
+    val cellType = {
+      val cellTypes = sources.map { source => Set(source.cellType) }.reduce { _ ++ _ }
```
echeipesh: This is totally a nit-pick but this pattern looks strange.
`val cellTypes = sources.map(_.cellType).toSet`
would be less surprising and probably a little more efficient since it could build a whole set at once rather than reducing.

vlm/src/main/scala/RasterSourceRDD.scala
```diff
@@ -0,0 +1,105 @@
...
+        throw new Exception(s"All RasterSources must be in the same projection, but multiple ones were found: $projections")
+      else
+        projections.head
+    }
+
+    val layerKeyBounds = {
+      val bounds =
+        sources.map { source =>
+          val projectedExtent = ProjectedExtent(source.extent, source.crs)
+          val boundsKey = projectedExtent.translate(SpatialKey(0, 0))
```
echeipesh: This part is confusing. The CRS of all the sources is required to be the same here and it must be the same CRS as the `LayoutDefinition`. So there is no need to reproject. And `SpatialKey(0,0)` looks like a placeholder.

You should union the source extents and use `KeyBounds(mapTransform(combinedExtents))` directly.

vlm/src/main/scala/RasterSourceRDD.scala
```diff
@@ -0,0 +1,105 @@
...
+
+    val sourcesRDD: RDD[(RasterSource, Array[RasterExtent])] =
+      sc.parallelize(sources).flatMap { source =>
+        val rasterExtents: Traversable[RasterExtent] =
+          extent.intersection(source.extent) match {
+            case Some(intersection) =>
+              val keys = mapTransform.keysForGeometry(intersection.toPolygon)
+
+              keys.map { key => RasterExtent(mapTransform(key), layout.tileCols, layout.tileRows) }
+            case None => Seq[RasterExtent]()
```
echeipesh: `Seq.empty[T]` or `Set.empty[T]` is more idiomatic

vlm/src/main/scala/RasterSourceRDD.scala
```diff
@@ -0,0 +1,105 @@
...
+    }
+
+    val layerMetadata =
+      TileLayerMetadata[SpatialKey](cellType, layout, combinedExtents, crs, layerKeyBounds)
+
+    val sourcesRDD: RDD[(RasterSource, Array[RasterExtent])] =
+      sc.parallelize(sources).flatMap { source =>
+        val rasterExtents: Traversable[RasterExtent] =
+          extent.intersection(source.extent) match {
+            case Some(intersection) =>
```
echeipesh: Just a note that you're basically manually translating from `Option` to `Seq`, and `option.toList.flatMap( intersection => ...)` is another way of expressing that. I'd say its a personal preference if `case match` or `toList.flatMap` is preferred. Mentioning it because it seems to come up a lot and is an interesting point.

vlm/src/main/scala/RasterSourceRDD.scala
```diff
@@ -0,0 +1,105 @@
...
+              keys.map { key => RasterExtent(mapTransform(key), layout.tileCols, layout.tileRows) }
+            case None => Seq[RasterExtent]()
+          }
+
+        if (rasterExtents.isEmpty)
+          Seq((source, rasterExtents.toArray))
+        else
+          RasterExtentPartitioner
+            .partitionRasterExtents(rasterExtents, partitionBytes)
+            .map { res => (source, res) }
```
echeipesh: The check of empty raster extents should be moved into `partitionRasterExtents` method. Its a always necessary check when calling this function right now and there is only one way to do it.

vlm/src/main/scala/TiledRasterRDD.scala
```diff
@@ -0,0 +1,24 @@
...
+import geotrellis.raster.reproject.Reproject
+import geotrellis.proj4._
+import geotrellis.spark._
+import geotrellis.spark.tiling._
+import org.apache.spark._
+import org.apache.spark.rdd._
+import cats.effect.IO
+
+
+object TiledRasterRDD {
```
echeipesh: Might as well remove these. These were mostly sketches I moving around.

echeipesh:
This is subset of changes from https://github.com/geotrellis/geotrellis-contrib/pull/8

PR #8, opened 2018-08-16
WarpRasterSource

jbouffard:
This PR adds the `WarpRasterSource` class that'll allow the user to reproject their `RasterSource`s into the desired projection.


vlm/src/main/scala/WarpRasterSource.scala
```diff
@@ -0,0 +1,58 @@
...
+  def cols: Int = base.cols
+  def rows: Int = base.rows
+  def cellType: CellType = base.cellType
+  def bandCount: Int = base.bandCount
+
+  def extent: Extent = {
+    val oldProjectedExtent = ProjectedExtent(base.extent, base.crs)
+
+    oldProjectedExtent.reprojectAsPolygon(crs).envelope
+  }
```
echeipesh: @jpolchlo Is this the correct way to do this or use of `ReprojectRasterExtent`?

vlm/src/main/scala/WarpRasterSource.scala
```diff
@@ -0,0 +1,58 @@
...
+  private val backTransform = Transform(crs, base.crs)
+
+  def read(windows: Traversable[RasterExtent]): Iterator[Raster[MultibandTile]] = {
+    val rasterExtents: Traversable[RasterExtent] =
+      windows.map { case targetRasterExtent =>
+        val projectedExtent = ProjectedExtent(targetRasterExtent.extent, crs)
+
+        val sourceExtent: Extent = projectedExtent.reprojectAsPolygon(base.crs).envelope
+
+        RasterExtent(sourceExtent, targetRasterExtent.cols, targetRasterExtent.rows)
```
echeipesh: This is not correct. We don't know how many pixels cover the same region in source raster vs target raster. You have to use the source `rasterExtent` to figure out what is the `GridBounds` of pixels corresponding to that extent.

vlm/src/main/scala/WarpRasterSource.scala
```diff
@@ -0,0 +1,58 @@
...
+      }
+
+    base.read(rasterExtents).map { raster =>
+      raster.reproject(base.crs, crs, options)
```
echeipesh: `options` should not be passed here. It has parameters that control how the selection of the target `RasterExtent` is controlled. Basically in this `reproject` call `errorThreshold`, `parentGridExtent`, and `targetRasterExtent`, `targetCellSize` can all produces different `targetRasterExtent`.

Instead of `ResampleMethod` should be exposed and the version of reproject that uses `targetRaterExtent` should be used.

vlm/src/main/scala/WarpRasterSource.scala
```diff
@@ -0,0 +1,58 @@
...
+  }
+}
+
+object WarpRasterSource {
+  def apply(
+    rasterSource: RasterSource,
+    crs: CRS,
+    options: Reproject.Options
+  ): WarpRasterSource =
+    new WarpRasterSource(rasterSource, crs, options)
```
echeipesh: It would save code and add clarity just to make this `case class` with default parameter of `resampleMethod = NeartestNeighbor`.

vlm/src/main/scala/WarpRasterSource.scala
```diff
@@ -0,0 +1,58 @@
...
+
+import geotrellis.proj4._
+import geotrellis.raster._
+import geotrellis.raster.reproject.{ReprojectRasterExtent, Reproject}
+import geotrellis.vector._
+
+import java.net.URI
+
+
+class WarpRasterSource(
```
echeipesh: This definitely needs a spec, probably something similar to `TileRDDReprojectSpec` where you compare reprojecting a whole raster that was produced using `WarpRasterSource`. It would probably be good idea to read 4 windows that divide the target raster and stitch them together to make sure that resampling is happening correctly around the tile borders.

vlm/src/main/scala/WarpRasterSource.scala
jbouffard: I think that `reproject` method should be fine since we're replacing the `options` parameter with the `resampleMethod`. By default, `targetRasterExtent`, `parentGridExtent`, and `targetCellSize` are all `None` https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/reproject/Reproject.scala#L38-L40 So the only thing that's going to be set is the `method`.

In addition, the version of the `reproject` method that does take a `targetRasterExtent` will use the default `Options` if none are passed in https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/reproject/RasterReprojectMethods.scala#L28 So I don't see how that version is better.

vlm/src/main/scala/WarpRasterSource.scala
```diff
@@ -0,0 +1,58 @@
...
+  val base: RasterSource,
+  val crs: CRS,
+  val options: Reproject.Options
+) extends RasterSource {
+
+  def uri: URI = base.uri
+  def cols: Int = base.cols
```
echeipesh: This is cols of the base raster, where we want cols in the reprojected raster.

PR #11, opened 2018-08-23
VLM: WarpGeoTiffRasterSource

echeipesh:
This PR introduces `WarpGeoTiffRasterSource` and `WarpSpec` that tests the fit of the reprojection in a more direct way. This is a general push to address and explore the following issues:

- The test success in more generic `WarpRasterSource` is contingent on grid of `targetRasterExtent` given to `read` function being the same as the default grid calculated by `ReprojectRasterExtent`. `WarpSpec` avoids that special case.
- In general reprjection can be seen as resample and warping in a single operation. When possible it is ideal to merge them into a single step to save a significant amount of computation and memory allocation. 

**Issues outstanding revealed:**
- `WarpGeoTiffRasterSource` does not pass `WarpSpec` when it uses full-tile reproject vs `RasterRegionReproject`. 
- It seems like there will likely have to be forced pairings of `Warp_` and `Source_` classes. Where GDAL source is forced to use `WarpedVRT`  and GeoTrellis `GeoTiffRasterSource` is forced to use `WarpGeoTiffRasterSource`

See: https://github.com/geotrellis/geotrellis-contrib/pull/11/commits/f668544c32e3bfe28687029ffa713de01cb6a7a5


vlm/src/main/scala/GeoTiffRasterSource.scala
```diff
@@ -11,31 +11,9 @@ import cats.effect.IO
...
-        throw new IllegalArgumentException(s"Unable to read scheme $scheme at $uri")
-    }
-    new StreamingByteReader(rr)
-  }
-
-  @transient private lazy val tiff: MultibandGeoTiff = GeoTiffReader.readMultiband(getByteReader, streaming = true)
-
-  def uri: URI = new URI(fileURI)
+case class GeoTiffRasterSource(uri: String) extends RasterSource {
+  @transient private lazy val tiff: MultibandGeoTiff =
```
jbouffard: How would you feel about having `GeoTiffInfo` being one of the properties of `GeoTiffRasterSource`? By having that, we'd have access to additional metadata like [bandCount](https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/io/geotiff/reader/GeoTiffReader.scala#L248) which could prove useful later on.

vlm/src/main/scala/Implicits.scala
```diff
@@ -13,32 +13,5 @@ object Implicits extends Implicits
 
 
 trait Implicits {
-  implicit def stringToURI(str: String): URI =
```
jbouffard: Do you think think that this implicit conversion won't be useful for `vlm`?

vlm/src/main/scala/WarpGeoTiffRasterSource.scala
```diff
@@ -0,0 +1,58 @@
...
+import geotrellis.raster.io.geotiff.MultibandGeoTiff
+import geotrellis.raster.io.geotiff.reader.GeoTiffReader
+import cats.effect.IO
+
+import java.net.URI
+import java.nio.file.Paths
+
+import geotrellis.util.{FileRangeReader, RangeReader, StreamingByteReader}
+
+case class WarpGeoTiffRasterSource(
```
jbouffard: I think it'd be nice to have just a single, generic `WarpRasterSource`; but I also think we'd have to decide on how we're going to implement the other features before deciding on the `WarpRasterSource` API.

vlm/src/main/scala/Implicits.scala
echeipesh: I think its best to avoid implicit conversions and overloads as much as possible while we're moving things around. They make it much harder to see what is going on and to refactor things.

vlm/src/main/scala/WarpGeoTiffRasterSource.scala
echeipesh: 👍

vlm/src/main/scala/GeoTiffRasterSource.scala
echeipesh: I'm pretty neutral about that. I would expect most of the usage of this class to actually happen through `RasterSource` interface rather than the concrete. `GeoTiffInfo` wouldn't be appropriate to expose in `RasterSource` because we expect it to abstract over GDAL. In general `GeoTiffInfo` has lots of useful metadata but I'm not sure how its going to end up being used.

jbouffard: Yeah, a lot of that metadata is GeoTiff specific, so I can see it having only some limited uses. What do you think about `RasterSource` having a `bandCount` member then? We already know that's an important piece of metadata, so it'd be good to have it available as soon as possible.

vlm/src/main/scala/Implicits.scala
jbouffard: :+1:

vlm/src/main/scala/GeoTiffRasterSource.scala
echeipesh: ‘bandCount’ is already on ‘RasterSource’

echeipesh: Not sure about the decision on VLM api, what do you mean? This mostly tries to address a direct problem with warp. I think things are going to become more clear once we make GDALRasterSource, that will need to lean on GDAL to perform the warp through virtual WRT.

Also you can see that reading band subsets will be desirable before long. This is incremental though.

jbouffard: Oh, sorry. I didn't realize that `bandCount` was already on `RasterSource`. I thought we were still debating on the API of `WarpRasterSource` and whether or not it is needed? I might have misunderstood the workflowy, though

PR #12, opened 2018-08-28
withCRS Method and Various Bug Fixes

jbouffard:
This PR changes the API of how `RasterSource`s are reprojected. Instead of using `WarpRasterSource` to do the reprojection, now `RasterSource`s have the `withCRS` method. This method will produce a new instance of the `RasterSource` that will read out tiles in the desired projection.

In addition to the API changes, various issues of the outputted tiles were resolved. Mainly, they will now be squares of the specified size.


vlm/src/main/scala/GeoTiffRasterSource.scala
```diff
@@ -44,47 +44,76 @@ class GeoTiffRasterSource(val fileURI: String) extends RasterSource {
   def bandCount: Int = tiff.bandCount
   def cellType: CellType = tiff.cellType
 
+  def withCRS(targetCRS: CRS, resampleMethod: ResampleMethod = NearestNeighbor): GeoTiffRasterSource =
+    new GeoTiffRasterSource(fileURI) {
```
echeipesh: Making anonymous classes like this should be avoided from code hygiene perspective. In the stack trace they're not going to have any meaningful name and it makes it not totally obvious where important code lives. This looks like `WarpGeoTiffRasterSource` from https://github.com/geotrellis/geotrellis-contrib/pull/11/files ? 

Either that PR should be merged and rebase or this class should get a name.

pomadchin: I think it can be merged as is, as after adding the unit test that compares this source to the common geotrellis tiling code it at least works as expected, after that we can combine it with your @echeipesh PR, what do you think?

echeipesh: That sounds good to me

pomadchin:
Merging, to hack on it along with https://github.com/geotrellis/geotrellis-contrib/pull/11

PR #18, opened 2018-09-12
[WIP] WarpGDALRasterSource Improvements

jbouffard:
This PR improves `WarpGDALRasterSource` by changing some of its eternal logic so that it uses more of the GDAL library and less of GeoTrellis. In addition, the `read` method will now work.

PR #20, opened 2018-09-20
VLM: RasterSource.read accepts Extent or GridBounds

echeipesh:
Old:
```scala
trait RasterSource {
  def rasterExtent: RasterExtent
  def read(windows: Traversable[RasterExtent]): Iterator[Raster[MultibandTile]]
}
```

New:
```scala
trait RasterSource {
  def rasterExtent: RasterExtent
  def read(extent: Extent, bands: Seq[Int]): Option[Raster[MultibandTile]]
  def read(bounds: GridBounds, bands: Seq[Int]): Option[Raster[MultibandTile]]
}
```

Reason:
From work on https://github.com/geotrellis/geotrellis-contrib/pull/18 where `RasterExtent` is backed by GDAL its clear that a new GDAL Dataset is required each time you change: resolution, grid-aligntment, CRS. The `Dataset` reference is a precious resource. Therefore its not proper to have a `read` method that potentially requires sources with multiple resolutions as `read(rasterExtent)` implies.

This also makes API more consistent, if `RasterExtent` has `rasterResolution` then its odd to be able to read windows with different resolutions. Changing things will probably be implemented with builder/mutator methods:

```scala
rasterSource
.withCRS(LatLng)
.withResampleMethod(NearestNeigbor)
.withTargetRasterExtent(re)
```


echeipesh:
Added test utils to actually interpret failing raster tests on the fly

screen shot 2018-09-25 at 11 26 30 am

PR #21, opened 2018-09-24
[WIP] GDAL Refactor

jbouffard:
This PR refactors the GDAL API such that it matches the API introduced in #20 In addition, this PR supersedes #18


vlm/src/main/scala/PaddedTile.scala
```diff
@@ -0,0 +1,12 @@
+package geotrellis.contrib.vlm
+
+
+import geotrellis.raster.GridBounds
+
+
+case class PaddedTile(
```
echeipesh: this should `extent Tile` so it can be returned as part of `MultibandTile` without changing any of the `RasterSource` signatures.

vlm/src/main/scala/RasterSource.scala
```diff
@@ -82,4 +82,8 @@ trait RasterSource extends Serializable {
     def readBounds(bounds: Traversable[GridBounds]): Iterator[Raster[MultibandTile]] =
         bounds.toIterator.flatMap(read(_, (0 until bandCount)).toIterator)
 
+    def readPaddedTiles(tiles: Traversable[PaddedTile], bands: Seq[Int]): Iterator[Raster[MultibandTile]]
```
echeipesh: We shouldn't add to the `RasterSource` interface like this. In the current formulation `PaddedTile` contains information about the pixel space from the `RasterSource` and some target layout. In general `RasterSource` should not be expected to handle data that is not directly related to the raster its sourcing.

jbouffard:
This PR is now very out-of-date and doesn't provide anything new, so I'm closing it.

PR #22, opened 2018-09-26
GDAL Raster Source 

pomadchin:
This PR adds a summary of a working part of the code in terms of the RasterSource GDAL support


vlm/src/main/scala/gdal/GDALDataType.scala
```diff
@@ -0,0 +1,40 @@
+package geotrellis.contrib.vlm.gdal
+
+import org.gdal.gdal.gdal
```
echeipesh: Best import ever

PR #23, opened 2018-09-26
Fix RasterSourceRDD square tile read

echeipesh:
- [x] Add `PaddedTile` class, this should allow us to make any tile "square"
- [x] Use `PaddedTile` to read out square tiles for RDD when there is not enough pixels in source data
- [x] Fix up most of the tests around RDD reading

One of the `RasterSourceRDD` tests is still failing but it is probably to do with resampling logic rather than source pixel alignment. The problem only happens at the edge of the partially overlapping tile. I think that can be investigated and fixed in different PR.


echeipesh:
Weirdly getting this current:
```scala
[error] /Users/eugene/proj/geotrellis-contrib/vlm/src/main/scala/PaddedTile.scala:22:12: class PaddedTile needs to be abstract, since:
[error] it has 4 unimplemented members.
[error] /** As seen from class PaddedTile, the missing signatures are as follows.
[error]  *  For convenience, these are usable as stub implementations.
[error]  */
[error]   // Members declared in geotrellis.macros.MacroIterableTile
[error]   def foreachDoubleVisitor(visitor: geotrellis.macros.DoubleTileVisitor): Unit = ???
[error]   def foreachIntVisitor(visitor: geotrellis.macros.IntTileVisitor): Unit = ???
[error]
[error]   // Members declared in geotrellis.macros.MacroMappableTile
[error]   def mapDoubleMapper(mapper: geotrellis.macros.DoubleTileMapper): geotrellis.raster.Tile = ???
[error]   def mapIntMapper(mapper: geotrellis.macros.IntTileMapper): geotrellis.raster.Tile = ???
[error] case class PaddedTile(chunk: Tile, colOffset: Int, rowOffset: Int, cols: Int, rows: Int) extends Tile {
[error]            ^
```

vlm/src/main/scala/LayoutTileSource.scala
```diff
@@ -0,0 +1,87 @@
...
+import geotrellis.spark.{SpatialKey}
+
+/** Reads tiles by key from a [[RasterSource]] as keyed by a [[LayoutDefinition]]
+  * @note It is required that the [[RasterSource]] is pixel aligned with the [[LayoutDefinition]]
+  *
+  * @param source raster source that can be queried by bounding box
+  * @param layout definition of a tile grid over the pixel grid
+  */
+class LayoutTileSource(val source: RasterSource, val layout: LayoutDefinition) {
+  // TODO: require that source and layout are grid alligned, consider floating point precision
```
jbouffard: Would it make sense then to add a `require` here that compares the two?

vlm/src/main/scala/LayoutTileSource.scala
```diff
@@ -0,0 +1,87 @@
...
+import geotrellis.spark.tiling._
+import geotrellis.spark.{SpatialKey}
+
+/** Reads tiles by key from a [[RasterSource]] as keyed by a [[LayoutDefinition]]
+  * @note It is required that the [[RasterSource]] is pixel aligned with the [[LayoutDefinition]]
+  *
+  * @param source raster source that can be queried by bounding box
+  * @param layout definition of a tile grid over the pixel grid
+  */
+class LayoutTileSource(val source: RasterSource, val layout: LayoutDefinition) {
```
jbouffard: Would it make sense to have a value or method that produces the keys which intersect the given `RasterSource`? That way the user knows which keys contain data and which ones don't.

vlm/src/main/scala/LayoutTileSource.scala
```diff
@@ -0,0 +1,87 @@
...
+      // raster is smaller but not bigger than I think ...
+      // its offset is relative to the raster we wished we had
+      val colOffset = bounds.colMin - sourcePixelBounds.colMin
+      val rowOffset = bounds.rowMin - sourcePixelBounds.rowMin
+      val tile = raster.tile.mapBands{ (_, band) =>
+        PaddedTile(band, colOffset, rowOffset, layout.tileCols, layout.tileRows)
+      }
+      (key, tile)
+    }
+  }
```
jbouffard: Going off my earlier comment, if we knew what keys intersected the `RasterSource` we could have a `readIntersectingKeys` method where the keys that are being read are the ones which intersect the `RasterSource`.

vlm/src/main/scala/LayoutTileSource.scala
```diff
@@ -0,0 +1,87 @@
...
+        PaddedTile(band, colOffset, rowOffset, layout.tileCols, layout.tileRows)
+      }
+    }
+  }
+
+  /** Read multiple tiles according to key.
+    * If each tile area intersects source partially the non-intersecting pixels will be filled with NODATA.
+    * If tile area does not intersect source it will be excluded from result iterator.
+    */
+  def readAll(keys: Iterator[SpatialKey]): Iterator[(SpatialKey, MultibandTile)] = {
```
jbouffard: Maybe this could have an overload where it'll just read all of the keys in the given `layout` maybe something like this?

```scala
def readAll(): Iterator[(SpatialKey, MultibandTile)] = {
  val arr = Array.ofDim[SpatialKey](layout.layoutCols * layout.layoutRows)
  val keys =
    for {
      col <- 0 until layout.layoutCols
      row <- 0 until layout.layoutRows
    } yield { SpatialKey(col, row) }

  readAll(keys)
}

File: vlm/src/main/scala/RasterSourceRDD.scala [hunk 5]
```diff
@@ -59,18 +59,17 @@ object RasterSourceRDD {
     val layerMetadata =
       TileLayerMetadata[SpatialKey](cellType, layout, combinedExtents, crs, layerKeyBounds)
 
-    val sourcesRDD: RDD[(RasterSource, Array[Extent])] =
+    val sourcesRDD: RDD[(RasterSource, Array[SpatialKey])] =
```
jbouffard: This is probably the best way to go, as I found that passing either `Extent`s or `GridBounds` can result in weird floating point issues.

vlm/src/main/scala/LayoutTileSource.scala
```diff
@@ -0,0 +1,87 @@
...
+  def sourceRowOffset: Long = ((layout.extent.ymax - source.extent.ymax) / layout.cellheight).toLong
+
+  /** Read tile according to key.
+    * If tile area intersects source partially the non-intersecting pixels will be filled with NODATA.
+    * If tile area does not intersect source None will be returned.
+    */
+  def read(key: SpatialKey): Option[MultibandTile] = {
+    val col = key.col.toLong
+    val row = key.row.toLong
+    val sourcePixelBounds = GridBounds(
```
jbouffard: Might be good to put this logic in its own private method.

pomadchin: Hm, i'd say `readAll` can be the following: 

```scala
def readAll(keys: Iterator[SpatialKey]): Iterator[(SpatialKey, MultibandTile)] =
    keys.flatMap(k => read(k).map(v => k -> v))
```

vlm/src/main/scala/LayoutTileSource.scala
echeipesh: Yes, totally. We should have something like `GridExtent.isAllignedWith(other: Gridextent)` but we don't have it yet. The concerns with putting `require` prematurely is that floating point precision can cause spurious failures that would be really frustrating.

vlm/src/main/scala/LayoutTileSource.scala
echeipesh: That sounds like it would be useful.

vlm/src/main/scala/LayoutTileSource.scala
echeipesh: You wouldn't have a key for every tile in the `LayoutDefintion`, pretty likely that each `RasterSource` is a small subset of overall coverage. You'd need to check which tiles the extent of `RasterSource` intersects.

vlm/src/main/scala/RasterSourceRDD.scala
echeipesh: Saw those as well. It's a little concerning because working from `Extent` should still be a first-class operation. But here working from the pixel/key bounds is a lot more direct anyway.

vlm/src/main/scala/LayoutTileSource.scala
echeipesh: The reason it doesn't is because our GeoTiffReader does a lot better when it reads multiple GridBounds at the same time, which is the method that would be invoked here. I don't know how long we can maintain the distinction between "read one" and "read multiple" in the API though.

vlm/src/main/scala/LayoutTileSource.scala
echeipesh: I think last commit should do it, but I'm prepared to see that code again.

vlm/src/main/scala/LayoutTileSource.scala
echeipesh: `def` is better than `val` btw so we don't have to hold on to the keys collection longer than its useful, its cheap enough to regenerate.

vlm/src/main/scala/LayoutTileSource.scala
echeipesh: `GridBounds` already has `offset` method that does this, just trying to save an extra class allocation here I guess?

PR #25, opened 2018-09-29
VLM: Introducing RasterRef

echeipesh:
First wack at this trying to re-assemble multiple bands before reading in the pixels.

Pretty light on comments and tests for the moment, just trying to drive it around.
Key design decisions:
- `RasterRef` is on the same pixel grid as the `RasterSource`
- `RasterRef` must not be contained by `RasterSource` bounds
  - this may/should make it easier to "move it around" and reproject as lazy transformations

`RasterRef` is `case class` for now but good design sense predicts that it should be a trait.
Basically being lazy right now and waiting to be forced into that eventuality.


pomadchin:
It looks like this PR is good to merge as the first step of a RasterSourceRefs support

PR #32, opened 2018-10-12
tileToLayout for RasterSource

### LINKED ISSUE #30: tileToLayout Method for RasterSource
# Overview
The current API around applying a layout to a `RasterSource` is a bit confusing. This is due to the fact that it's not clear if `LayoutTileSource` fits its `RasterSource` to the given layout, or if it applies the layout to the `RasterSource`. As of right now, it does former. However, this differs from GeoTrellis API where the latter is performed through methods like `tileToLayout`.

## Background

To help illustrate these differences, two images are provided below. In both cases, the image is tiled to a layout that has 2 `layoutCols` and 2 `layoutRows`.

### Fitting an Image Into a Layout

layout_2

The above picture shows what is currently done in `LayoutTileSource`. Where a given `RasterSource` is fitted into the supplied layout.

### Applying a Layout to an Image

layout_1

In the second example, the layout is applied to the image. This is what is done in GeoTrellis when using the `tileToLayout` method.

## Solution

A solution to this issue would be to add a `tileToLayout` method to `RasterSource`. This will still produce a `LayoutTileSource`, but its layout will be formatted such that it

### Example API

```scala
def tileToLayout(layout: LayoutDefinition): LayoutTileSource = ???

val rasterSource: RasterSource = ???
val targetLayout: LayoutDefinition = ???

val tiledRasterSource: LayoutTileSource = rasterSource.tileToLayout(targetLayout)
```

jbouffard:
This PR adds the `tileToLayout` method to `RasterSource` which will allow users to apply a layout on their source data.

Resolves #30

### DIFF
--- a/vlm/src/test/scala/geotrellis/contrib/vlm/GeoTiffRasterSourceSpec.scala
+++ b/vlm/src/test/scala/geotrellis/contrib/vlm/GeoTiffRasterSourceSpec.scala
@@ -17,12 +17,20 @@
 package geotrellis.contrib.vlm
 
 import geotrellis.raster._
+import geotrellis.raster.prototype._
+import geotrellis.raster.merge._
+import geotrellis.raster.io.geotiff.MultibandGeoTiff
 import geotrellis.raster.io.geotiff.reader.GeoTiffReader
 import geotrellis.raster.resample._
 import geotrellis.raster.reproject._
 import geotrellis.raster.testkit._
+import geotrellis.vector.{Extent, ProjectedExtent}
 import geotrellis.proj4._
+import geotrellis.spark._
+import geotrellis.spark.tiling._
 import geotrellis.spark.testkit._
+import geotrellis.util._
+
 import org.scalatest._
 
 import java.io.File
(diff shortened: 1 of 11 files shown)


vlm/src/test/scala/geotrellis/contrib/vlm/GeoTiffRasterSourceSpec.scala
```diff
@@ -73,4 +73,88 @@ class GeoTiffRasterSourceSpec extends FunSpec with RasterMatchers with BetterRas
       assertRastersEqual(actual, expected)
     }
   }
+
+  it("should perform a tileToLayout") {
+    val targetCellSize: CellSize = CellSize(20.0, 20.0)
+
+    val testSource = GeoTiffReader.readMultiband(url).getClosestOverview(targetCellSize)
```
pomadchin: There is a questions of a separate PR which i am working on, why the same test without selecting on overview works for GDAL?

vlm/src/test/scala/geotrellis/contrib/vlm/gdal/GDALRasterSourceSpec.scala
```diff
@@ -75,4 +81,66 @@ class GDALRasterSourceSpec extends FunSpec with RasterMatchers with BetterRaster
...
+          sortedActual.zip(sortedExpected)
+
+        grouped.foreach { case (actualTile, expectedTile) =>
+          withGeoTiffClue(actualTile, expectedTile, source.crs) {
+            assertRastersEqual(actualTile, expectedTile)
+          }
+        }
+
+        resampledSource.close
+        layoutSource.source.close
```
pomadchin: @jbouffard @echeipesh the first time i actually had to clean up memory, without it i had:

```
error] Error during tests:
[error] 	Running java with options -classpath ... <the huge classpath there> ...
```

pomadchin:
Merging!

vlm/src/test/scala/geotrellis/contrib/vlm/gdal/GDALRasterSourceSpec.scala
echeipesh: Was there a useful message with this error or just that ?

PR #34, opened 2018-10-16
0.2.0 Release

jbouffard:
This PR bumps the version number of geotrellis vlm from 0.1.1 to 0.2.0

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -55,7 +55,7 @@ lazy val vlm = project
   .settings(
     organization := "com.azavea.geotrellis",
     name := "geotrellis-contrib-vlm",
-    version := "0.1.1",
+    version := "0.2.0",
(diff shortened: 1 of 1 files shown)


jbouffard:
Hmmm..... What do you think @echeipesh? I'm fine with doing it either way.

pomadchin:
mm lets merge it as it is, fork it and update version up to `-SNAPSHOT`?

pomadchin:
https://github.com/geotrellis/geotrellis-contrib/tree/0.2.0

echeipesh:
Can't publish `-SNAPSHOT` to bintray. But bumping the version after publish is a really good practice. In theory it should happen with the next PR that comes in, but probably really easy to lose track of that, so publish, bump version looks like more realistic workflow for the future.

PR #39, opened 2018-10-23
VLM: RasterSourceMetadata collection

pomadchin:
Thie PR adds metadata collection and conversion for the VLM API

### DIFF
--- a/vlm/src/main/scala/geotrellis/contrib/vlm/GeoTiffReprojectRasterSource.scala
+++ b/vlm/src/main/scala/geotrellis/contrib/vlm/GeoTiffReprojectRasterSource.scala
@@ -94,9 +94,7 @@ class GeoTiffReprojectRasterSource(
     new GeoTiffReprojectRasterSource(uri, targetCRS, options)
 
   def resample(resampleGrid: ResampleGrid, method: ResampleMethod): RasterSource =
-    new GeoTiffResampleRasterSource(uri, resampleGrid, method) {
-      override def crs: CRS = self.crs
+    new GeoTiffReprojectRasterSource(uri, crs, options.copy(method = method)) {
       override lazy val rasterExtent = resampleGrid(self.rasterExtent)
     }
-
 }
--- a/vlm/src/main/scala/geotrellis/contrib/vlm/GeoTiffReprojectRasterSource.scala
+++ b/vlm/src/main/scala/geotrellis/contrib/vlm/GeoTiffReprojectRasterSource.scala
@@ -94,7 +94,8 @@ class GeoTiffReprojectRasterSource(
     new GeoTiffReprojectRasterSource(uri, targetCRS, options)
 
   def resample(resampleGrid: ResampleGrid, method: ResampleMethod): RasterSource =
-    new GeoTiffReprojectRasterSource(uri, crs, options.copy(method = method)) {
-      override lazy val rasterExtent = resampleGrid(self.rasterExtent)
-    }
+    new GeoTiffReprojectRasterSource(uri, crs, options.copy(
+      method = method,
+      targetRasterExtent = Some(resampleGrid(self.rasterExtent))
+    ))
 }
--- a/vlm/src/main/scala/geotrellis/contrib/vlm/GeoTiffReprojectRasterSource.scala
+++ b/vlm/src/main/scala/geotrellis/contrib/vlm/GeoTiffReprojectRasterSource.scala
@@ -29,14 +29,15 @@ class GeoTiffReprojectRasterSource(
   val crs: CRS,
   val options: Reproject.Options = Reproject.Options.DEFAULT
 ) extends RasterSource { self =>
-  @transient lazy val tiff: MultibandGeoTiff =
+  @transient protected lazy val tiff: MultibandGeoTiff =
     GeoTiffReader.readMultiband(getByteReader(uri), streaming = true)
 
-  lazy val baseCRS = tiff.crs
-  lazy val baseRasterExtent = tiff.rasterExtent
+  protected lazy val baseCRS = tiff.crs
+  protected lazy val baseRasterExtent = tiff.rasterExtent
 
-  lazy val transform = Transform(baseCRS, crs)
-  lazy val backTransform = Transform(crs, baseCRS)
+  protected lazy val transform = Transform(baseCRS, crs)
+  protected lazy val backTransform = Transform(crs, baseCRS)
+
   override lazy val rasterExtent: RasterExtent = options.targetRasterExtent match {
     case Some(targetRasterExtent) =>
       targetRasterExtent
--- a/vlm/src/main/scala/geotrellis/contrib/vlm/GeoTiffReprojectRasterSource.scala
+++ b/vlm/src/main/scala/geotrellis/contrib/vlm/GeoTiffReprojectRasterSource.scala
@@ -28,15 +28,15 @@ class GeoTiffReprojectRasterSource(
   val uri: String,
   val crs: CRS,
   val options: Reproject.Options = Reproject.Options.DEFAULT
-) extends RasterSource {
-  @transient private lazy val tiff: MultibandGeoTiff =
+) extends RasterSource { self =>
+  @transient lazy val tiff: MultibandGeoTiff =
     GeoTiffReader.readMultiband(getByteReader(uri), streaming = true)
 
-  private lazy val baseCRS = tiff.crs
-  private lazy val baseRasterExtent = tiff.rasterExtent
+  lazy val baseCRS = tiff.crs
+  lazy val baseRasterExtent = tiff.rasterExtent
 
-  private lazy val transform = Transform(baseCRS, crs)
-  private lazy val backTransform = Transform(crs, baseCRS)
+  lazy val transform = Transform(baseCRS, crs)
+  lazy val backTransform = Transform(crs, baseCRS)
   override lazy val rasterExtent: RasterExtent = options.targetRasterExtent match {
     case Some(targetRasterExtent) =>
       targetRasterExtent
@@ -94,6 +94,9 @@ class GeoTiffReprojectRasterSource(
     new GeoTiffReprojectRasterSource(uri, targetCRS, options)
 
   def resample(resampleGrid: ResampleGrid, method: ResampleMethod): RasterSource =
-    new GeoTiffResampleRasterSource(uri, resampleGrid, method)
+    new GeoTiffResampleRasterSource(uri, resampleGrid, method) {
+      override def crs: CRS = self.crs
+      override lazy val rasterExtent = resampleGrid(self.rasterExtent)
+    }
 
 }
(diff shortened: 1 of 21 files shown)


vlm/src/main/scala/geotrellis/contrib/vlm/GeoTiffReprojectRasterSource.scala
```diff
@@ -94,6 +94,9 @@ class GeoTiffReprojectRasterSource(
     new GeoTiffReprojectRasterSource(uri, targetCRS, options)
 
   def resample(resampleGrid: ResampleGrid, method: ResampleMethod): RasterSource =
-    new GeoTiffResampleRasterSource(uri, resampleGrid, method)
+    new GeoTiffResampleRasterSource(uri, resampleGrid, method) {
```
pomadchin: Thats why i had metadata mismatch. In fact chained operations didn't work. Chained means `RasterSource.reproject.resample... etc`

echeipesh: `GeoTiffReprojectRasterSource!.resample` should produce `GeoTiffReprojectRasterSource` with modified rasterExtent, not `GeoTiffResampleRasterSource`

vlm/src/test/scala/geotrellis/contrib/vlm/RasterSummarySpec.scala
```diff
@@ -0,0 +1,215 @@
...
+      widthCollected shouldBe (widthResampled +- 1e-7)
+      heightCollected shouldBe (heightResampled +- 1e-7)
+
+      summaryCollected.extent shouldBe summaryResampled.extent
+      summaryCollected.cells shouldBe summaryResampled.cells
+      summaryCollected.count shouldBe summaryResampled.count
+    }
+  }
+
+  it("should create ContextRDD from RDD of RasterSources") {
```
pomadchin: that's actually what @jbouffard needs

vlm/src/test/scala/geotrellis/contrib/vlm/RasterSummarySpec.scala
```diff
@@ -0,0 +1,215 @@
...
+
+      rasterSource.crs shouldBe summary.crs
+      rasterSource.extent shouldBe summary.extent
+      rasterSource.cellSize shouldBe summary.cellSize
+      rasterSource.cellType shouldBe summary.cellType
+      rasterSource.size shouldBe summary.cells
+      files.length shouldBe summary.count
+    }
+
+    // TODO: the problem is in a GDAL -tap parameter usage
```
pomadchin: this is the `-tap` parameter usage issue

vlm/src/test/scala/geotrellis/contrib/vlm/RasterSummarySpec.scala
```diff
@@ -0,0 +1,215 @@
...
+
+      summaryCollected.cells shouldBe summaryResampled.cells
+      summaryCollected.count shouldBe summaryResampled.count
+    }
+  }
+
+  // TODO: fix JNI management here
+  // this test works as expected, though crashes with a core dump exception
+  // works slower than a GeoTiff version, there is an interesting discussion here: https://github.com/OpenDroneMap/OpenDroneMap/issues/778
+  // also mb we should pass a resampled RasterExtent like we did in a GeoTiff case
```
pomadchin: This test can work only after the class path cleanup

vlm/src/test/scala/geotrellis/contrib/vlm/RasterSummarySpec.scala
jbouffard: :+1:

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSummary.scala
```diff
@@ -0,0 +1,100 @@
...
+import geotrellis.util._
+import geotrellis.vector.{Extent, ProjectedExtent}
+import org.apache.spark.rdd.RDD
+
+case class RasterSummary(
+  crs: CRS,
+  cellType: CellType,
+  cellSize: CellSize,
+  extent: Extent,
+  cells: Long,
```
jbouffard: This doesn't need to happen for this PR, but based on our discussion earlier, how would you feel about `RasterSummary` having both `GridBounds` and the number of non-`NoData` values in a given `RasterSource`?

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSummary.scala
```diff
@@ -0,0 +1,100 @@
...
+}
+
+object RasterSummary {
+  /** Collect [[RasterSummary]] from unstructred rasters, grouped by CRS */
+  def collect[V <: CellGrid: GetComponent[?, ProjectedExtent]](rdd: RDD[V]): Seq[RasterSummary] = {
+    rdd
+      .map { grid =>
+        val ProjectedExtent(extent, crs) = grid.getComponent[ProjectedExtent]
+        val cellSize = CellSize(extent, grid.cols, grid.rows)
+        (crs, RasterSummary(crs, grid.cellType, cellSize, extent, grid.size, 1))
```
jbouffard: @echeipesh Did we want the number of non-`NoData` values for this PR? Or can that be pushed off until later?

vlm/src/main/scala/geotrellis/contrib/vlm/GeoTiffReprojectRasterSource.scala
```diff
@@ -28,15 +28,15 @@ class GeoTiffReprojectRasterSource(
...
   val crs: CRS,
   val options: Reproject.Options = Reproject.Options.DEFAULT
-) extends RasterSource {
-  @transient private lazy val tiff: MultibandGeoTiff =
+) extends RasterSource { self =>
+  @transient lazy val tiff: MultibandGeoTiff =
     GeoTiffReader.readMultiband(getByteReader(uri), streaming = true)
 
-  private lazy val baseCRS = tiff.crs
-  private lazy val baseRasterExtent = tiff.rasterExtent
```
echeipesh: Why remove all the `private` ? Not a problem now but eventually I could see this becoming part of the API unintentionally.

vlm/src/main/scala/geotrellis/contrib/vlm/gdal/GDALBaseRasterSource.scala
```diff
@@ -6,11 +6,14 @@ import geotrellis.raster._
 import geotrellis.raster.reproject.Reproject
 import geotrellis.raster.resample.ResampleMethod
 import geotrellis.vector._
+
 import org.gdal.gdal.{Dataset, gdal}
 import org.gdal.osr.SpatialReference
 
 trait GDALBaseRasterSource extends RasterSource {
+  val uri: String
```
jbouffard: Why did you add `uri` here as a `val` when `RasterSource` already has it as a method?

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSummary.scala
```diff
@@ -0,0 +1,100 @@
...
+  def toTileLayerMetadata(layoutDefinition: LayoutDefinition, zoom: Int): (TileLayerMetadata[SpatialKey], Some[Int]) =
+    toTileLayerMetadata(LayoutLevel(zoom, layoutDefinition))
+
+  def toTileLayerMetadata(layoutType: LayoutType): (TileLayerMetadata[SpatialKey], Option[Int]) = {
+    val (ld, zoom) = layoutType.layoutDefinitionWithZoom(crs, extent, cellSize)
+    val dataBounds: Bounds[SpatialKey] = KeyBounds(ld.mapTransform.extentToBounds(extent))
+    TileLayerMetadata[SpatialKey](cellType, ld, extent, crs, dataBounds) -> zoom
+  }
+
+  def resample(resampleGrid: ResampleGrid): RasterSummary = {
```
jbouffard: Should this be moved to `RasterSummarySpec`? Since earlier we talked about how it doesn't make much sense for `RasterSummary` to have a `resample` method.

pomadchin: I think this function should still be there, as it can be very useful for the test purposes, at least for now.

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSummary.scala
pomadchin: At this point it's not resolved actually :/ I'd like to resolve in a separate PR. At the moment cells is just a `cols x rows` value.

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSummary.scala
pomadchin: What's wrong with it? GridBounds are the actual bounds and number of NoData values is a `cells` number. This value is used to estimate better partition sizes as you remember from our conversation. At this point `cells` just computed a bit incorrect, see the comment below.

vlm/src/main/scala/geotrellis/contrib/vlm/GeoTiffReprojectRasterSource.scala
pomadchin: Ah will fix it, forgot to revert changes after the debug process.

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSummary.scala
jbouffard: Oh, I misspoke. I meant that it might be useful to have the number of all pixels and the number of non-`NoData` values

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSummary.scala
jbouffard: That sounds good to me :+1:

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSummary.scala
jbouffard: That makes sense. I think we can move it to to the tests at some point instead of removing it.

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSummary.scala
pomadchin: Number of all pixels would be GridBounds size.

jbouffard: Ah, that's true.

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSummary.scala
pomadchin: @jbouffard that's true I added a TODO thing, probably i will remove it in a separate PR!

vlm/src/test/scala/geotrellis/contrib/vlm/RasterSummarySpec.scala
pomadchin: the dump contains the following: 

```
# Problematic frame:
# C  [libgdal.20.dylib+0x4a430e]  GDALOverviewBand::FlushCache()+0x14
```

PR #42, opened 2018-11-01
Fix GDAL JNI references management

### LINKED ISSUE #36: Free GDAL related resources
We can consider https://github.com/jsuereth/scala-arm library usage as suggested @echeipesh to work better with resources that should be closed.

pomadchin:
This PR fixes bad GDALRasterSources JNI references management.

Fixes #36


vlm/src/test/scala/geotrellis/contrib/vlm/RasterSourceRDDSpec.scala
```diff
@@ -146,26 +146,33 @@ class RasterSourceRDDSpec extends FunSpec with TestEnvironment with BetterRaster
 
     describe("GDALRasterSource") {
       val expectedFilePath = s"${new File("").getAbsolutePath()}/src/test/resources/img/aspect-tiled-near-merc-rdd.tif"
-      val reprojectedExpectedRDDGDAL: MultibandTileLayerRDD[SpatialKey] =
-        RasterSourceRDD(GDALRasterSource(expectedFilePath), layout)
 
-      val rasterSource = GDALRasterSource(filePath)
       it("should reproduce tileToLayout") {
+        val rasterSource = GDALRasterSource(filePath)
```
jbouffard: This isn't an issue so much as a question, but will we need to instantiate a `GDALRasterSource` for each test from now one?

vlm/src/main/scala/geotrellis/contrib/vlm/gdal/GDALResampleRasterSource.scala
```diff
@@ -8,59 +8,53 @@ import geotrellis.proj4.CRS
...
+  method: ResampleMethod = NearestNeighbor,
+  baseWarpList: List[GDALWarpOptions] = Nil
+) extends GDALBaseRasterSource {
+  lazy val warpOptions: GDALWarpOptions = {
+    val res = resampleGrid match {
+      case Dimensions(cols, rows) =>
+        GDALWarpOptions(
+          dimensions = (cols, rows).some,
+          resampleMethod = method.some
+        )
```
jbouffard: Should there be a `case` that matches for [TargetRegion](https://github.com/geotrellis/geotrellis-contrib/blob/master/vlm/src/main/scala/geotrellis/contrib/vlm/ResampleGrid.scala#L35) here? Since we'll just be using the `RasterExtent` that object has anyway.

vlm/src/main/scala/geotrellis/contrib/vlm/gdal/GDALResampleRasterSource.scala
```diff
@@ -8,59 +8,53 @@ import geotrellis.proj4.CRS
...
+
+          val geoTransform: Array[Double] = baseDataset.GetGeoTransform
+
+          val xmin: Double = geoTransform(0)
+          val ymin: Double = geoTransform(3) + geoTransform(5) * rows
+          val xmax: Double = geoTransform(0) + geoTransform(1) * cols
+          val ymax: Double = geoTransform(3)
+          baseDataset.delete()
+
+          RasterExtent(Extent(xmin, ymin, xmax, ymax), cols, rows)
```
jbouffard: In the case of `TargetGrid`, I don't think we need a `RasterExtent`. Since `TargetGrid.apply` [just uses the Extent](https://github.com/geotrellis/geotrellis-contrib/blob/master/vlm/src/main/scala/geotrellis/contrib/vlm/ResampleGrid.scala#L32).

vlm/src/test/scala/geotrellis/contrib/vlm/RasterSourceRDDSpec.scala
pomadchin: Yes, just to be sure that we clean links as expected and in a proper moment, if we'll use the same resource for multiple tests, than there would appear a question, when do we want to close it?

vlm/src/main/scala/geotrellis/contrib/vlm/gdal/GDALResampleRasterSource.scala
pomadchin: No, TargetRegion.apply returns the initial RasterExtent. I decided not to copy-paste the code.

vlm/src/main/scala/geotrellis/contrib/vlm/gdal/GDALResampleRasterSource.scala
pomadchin: `.apply` uses [RasterExtent](https://github.com/geotrellis/geotrellis-contrib/blob/master/vlm/src/main/scala/geotrellis/contrib/vlm/ResampleGrid.scala#L31) It's a generic interface to make a generic apply for all three `ResampleGrid` types.

vlm/src/main/scala/geotrellis/contrib/vlm/RasterRegion.scala
```diff
@@ -53,4 +53,5 @@ case class RasterRegion(
   def cellType: CellType = source.cellType
   def rasterExtent: RasterExtent = RasterExtent(extent, cols, rows)
   def projectedExtent: ProjectedExtent = ProjectedExtent(extent, crs)
+  def close = source.close
```
echeipesh: This doesn't seem right. There could be multiple `RasterRegions` that are floating and have lifetime independent of `RasterSource`. Closing the source here seems like it would be problematic, is there a reason this ended up here?

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSource.scala
```diff
@@ -42,7 +40,7 @@ import java.io.Closeable
   * @groupdesc reproject Functions to resample raster data in target projection.
   * @groupprio reproject 2
   */
-trait RasterSource extends CellGrid with Closeable with Serializable {
+trait RasterSource extends CellGrid with AutoCloseable with Serializable {
```
echeipesh: Whats the practical difference between `Closable` and `AutoClosable` why was this switch helpful ?

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSummary.scala
```diff
@@ -15,13 +31,13 @@ case class RasterSummary(
   extent: Extent,
   cells: Long,
   count: Long
-) {
+) extends LazyLogging {
```
echeipesh: An aside: how do you feel about this pattern? 
I'm not sure if its "bad" but it feels weird to include this in type definition just to get a private field for `logger`.

vlm/src/main/scala/geotrellis/contrib/vlm/ResampleGrid.scala
```diff
@@ -19,20 +19,21 @@ package geotrellis.contrib.vlm
 import geotrellis.raster.{RasterExtent, GridExtent}
 
 sealed trait ResampleGrid {
-  def apply(source: RasterExtent): RasterExtent
+  // this is a by name parameter, as we don't need to call the source in all ResampleGrid types
+  def apply(source: => RasterExtent): RasterExtent
```
echeipesh: 👍

vlm/src/main/scala/geotrellis/contrib/vlm/cache/LazyCache.scala
```diff
@@ -0,0 +1,58 @@
...
+ * limitations under the License.
+ */
+
+package geotrellis.contrib.vlm.cache
+
+import com.github.blemale.scaffeine.Cache
+import scala.collection.concurrent.TrieMap
+
+/** A limited cache wrapper that can be "disable" caching (lazy instance creation) */
+case class LazyCache[K, V](getUnderlying: () => Cache[K, V], enabled: Boolean = true) {
```
echeipesh: Is it lazy or just configuration wrapper? Name is just weird.

vlm/src/main/scala/geotrellis/contrib/vlm/gdal/GDALBaseRasterSource.scala
```diff
@@ -114,8 +134,10 @@ trait GDALBaseRasterSource extends RasterSource {
     readBounds(bounds, 0 until bandCount)
   }
 
-  override def close = {
-    // baseDataset.delete()
-    dataset.delete()
-  }
+  override def close = dataset.delete()
+
+  /*override def finalize(): Unit = {
```
echeipesh: why did this end up commented out? If not needed why not delete ?

vlm/src/main/scala/geotrellis/contrib/vlm/RasterRegion.scala
pomadchin: Smth that should not be there; legacy from lots of test iterations.

vlm/src/main/scala/geotrellis/contrib/vlm/cache/LazyCache.scala
pomadchin: It's lazy, there would be no cache instance created if that would be create with a param `enabled = false`. Hm, do you think smth like `DisableableCache` sounds better? O:

vlm/src/main/scala/geotrellis/contrib/vlm/gdal/GDALBaseRasterSource.scala
pomadchin: Should be removed!

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSource.scala
pomadchin: So the `Closeable` is a legacy type with a more restricted exception type (it can throw only `IOException`). `AutoCloseable` can be used with Java >= 1.7 only.

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSummary.scala
pomadchin: Well, i dont feel really bad because of it; as its what is suggested by scalalogging: https://github.com/lightbend/scala-logging/blob/d271c40599bd6cc4dcee35b90768506f50b26e33/src/main/scala/com/typesafe/scalalogging/Logging.scala#L25 

Also it looks like from 3.9.0 it is transient, so we can remove this trait in 3.0!

PR #43, opened 2018-11-06
Passing in URIs to Gdal and GdalReader

jbouffard:
# Overview

This PR started off as a copy of another PR: https://github.com/geotrellis/geotrellis-gdal/pull/20 However, it has since diverged from that source to become something of its own.

This PR makes it so that users can pass in paths to `GDAL` and `GDALReader`. This was done by creating the `VSIPath` case class which formats the given path in such a way that GDAL can read it.

The backends that are supported with this PR:
- Local Filesystem
- Hdfs
- S3
- Azure
- Google Cloud Storage
- HTTP/HTTPS/FTP

In additions to the supported backends, various compressed filetypes can be read from as well:
- zip
- kmz
- ods
- xlsx
- gzip
- tar
- tgz

When reading a compressed file, there's two different ways one can format their path:
- `s3://bucket/path/data.zip`
- `zip+s2://bucket/path/data.zip`

It is also possible to read a file from a compressed directory. The format reading such a file is can be:
- `s3://bucket/path/data.zip!file.tiff`
- `zip+s3://bucket/path/data.zip!file.tiff`

## Caveats and Known Issues

- The Window's cases have not been tests due to not having access to a Window's machine.
- It's not possible to chain together more than two handlers. So for example, you can't read a file that's in a zip file that's in another zip file that's on S3.
- There might be certain paths that contain a combination of characters such that `VSIPath` cannot parse it.


vlm/src/main/scala/geotrellis/contrib/vlm/gdal/URIFormatter.scala
```diff
@@ -0,0 +1,91 @@
...
+          case Some(scheme) => s"/vsi$scheme/"
+          case None => ""
+        }
+      else
+        ""
+
+    val formattedPath: String =
+      uri.getScheme match {
+        case (FTP | HTTP | HTTPS) =>
+          s"/vsicurl/${uri.toString}"
```
pomadchin: Is not `uri.toString` already a `path`?

vlm/src/main/scala/geotrellis/contrib/vlm/gdal/URIFormatter.scala
```diff
@@ -0,0 +1,91 @@
...
+    // TODO: Support reading a single file from a zip file
+    // ie. reading 1.tiff from a zip: s3://bucket/path/to/my/zipped/data/data.zip/1.tiff
+
+    val path = uri.toString
+
+    // We must first determine if we need to chain file handlers
+    // ie. Reading from a zip file on s3
+    val leadingScheme: String =
+      if (path.contains("."))
+        FILE_TYPE_TO_SCHEME.get(path.substring(path.lastIndexOf(".") + 1)) match {
```
pomadchin: can it handle smth [like that](https://github.com/mapbox/rasterio/blob/960a906dad2a4e426387ce048a52c6e90afdcd2b/rasterio/path.py#L45) ?

vlm/src/main/scala/geotrellis/contrib/vlm/gdal/URIFormatter.scala
```diff
@@ -0,0 +1,91 @@
+package geotrellis.contrib.vlm.gdal
+
+
+import java.net.URI
+
+
+object Schemes {
+  final val FTP = "ftp"
```
pomadchin: Hm, the only way to resolve scheme is pattern matching?

vlm/src/test/scala/geotrellis/contrib/vlm/gdal/URIFormatterSpec.scala
```diff
@@ -0,0 +1,73 @@
+package geotrellis.contrib.vlm.gdal
+
+import java.net.URI
+
+import org.scalatest._
+
+
+class URIFormatterSpec extends FunSpec with Matchers {
```
pomadchin: Can we add more test cases here? https://github.com/mapbox/rasterio/blob/master/tests/test_path.py

vlm/src/main/scala/geotrellis/contrib/vlm/gdal/URIFormatter.scala
```diff
@@ -0,0 +1,91 @@
...
+      ZIP -> ZIP,
+      "kmz" -> ZIP,
+      "ods" -> ZIP,
+      "xlsx" -> ZIP,
+      GZIP -> GZIP
+    )
+}
+
+
+object URIFormatter {
```
echeipesh: Can i quibble about the name a bit? ‘URIformatter’ could mean almost anything.

Some alternatives are: ‘UriVsiFormatter’, package function named ‘uri2vsi’ or ‘VsiPath’ that has ‘fromUri’ and ‘toUri’ methods.

pomadchin: My vote would be for `VSIPath`, and we can wrap `URI` into it (smth like what was implemented in rasterio)

vlm/src/main/scala/geotrellis/contrib/vlm/gdal/GDAL.scala
```diff
@@ -88,8 +90,11 @@ object GDAL {
       case _ => throw new Exception(s"Could not find equivalent GDALResampleMethod for: $method")
     }
 
-  def open(path: String): Dataset = {
-    val ds = gdal.Open(path, gdalconstConstants.GA_ReadOnly)
+  def open(path: String): Dataset =
+    open(new URI(path))
+
+  def open(uri: URI): Dataset = {
```
echeipesh: @pomadchin do you remember exactly why we had to avoid the ‘URI’ container? https://github.com/raster-foundry/raster-foundry/blob/develop/app-backend/common/src/main/scala/utils/RangeReaderUtils.scala Something about a wacky s3 uri violated the character constraints and I am not sure if it would be an issue here, generating vsipath from such uri

pomadchin: @echeipesh i don't remember actually, but yeah there was smth with it, I'll double check with James about RF case

pomadchin: @echeipesh 
```
James:
one possible reason -- the `%7C` / `|` thing because we're using auth0 users for a component of the path
```

^ that's definitely smth i remember was a problem in RF, and it was related to how `URI` handles special characters.

vlm/src/main/scala/geotrellis/contrib/vlm/gdal/VSIPath.scala
```diff
@@ -0,0 +1,150 @@
...
+        case Some(scheme) => s"/vsi$scheme/"
+        case None => ""
+      }
+
+    def formatPath(targetURI: URI): String = {
+      lazy val uriPath = targetURI.getPath
+
+      targetURI.getScheme match {
+        case (FTP | HTTP | HTTPS) =>
+          s"/vsicurl/${targetURI.toString}"
```
pomadchin: `"${targetURI}"` would be enough there, it calls `toString` to convert `targetURI` into `String` itself.

vlm/src/test/scala/geotrellis/contrib/vlm/gdal/VSIPathSpec.scala
```diff
@@ -0,0 +1,136 @@
...
+    it("should format - Azure uri") {
+      val uri = "wasb://test-files@myaccount.blah.core.net/nlcd/data/tiff-0.tiff"
+      val expectedPath = "/vsiaz/test-files/nlcd/data/tiff-0.tiff"
+
+      VSIPath(uri) should be (expectedPath)
+    }
+
+    it("should format - chained Azure uri") {
+      val uri = "wasb://test-files@myaccount.blah.core.net/nlcd/data/info.kmz"
+      val expectedPath = "/vsizip//vsiaz/test-files/nlcd/data/info.kmz"
```
pomadchin: Double slash there `/vsizip//vsiaz/` => should be `/vsizip/vsiaz/`

vlm/src/main/scala/geotrellis/contrib/vlm/gdal/VSIPath.scala
```diff
@@ -0,0 +1,150 @@
...
+      ZIP -> ZIP,
+      "kmz" -> ZIP,
+      "ods" -> ZIP,
+      "xlsx" -> ZIP,
+      GZIP -> GZIP
+    )
+}
+
+
+object VSIPath {
```
pomadchin: Hm, do we want to have `VSIPath` as a function (the current implementation) or we want to have some class (similar to `URI`)? (It's just a question, and both options are good to me)

vlm/src/main/scala/geotrellis/contrib/vlm/gdal/VSIPath.scala
```diff
@@ -0,0 +1,150 @@
...
+        case (WASB | WASBS) =>
+          val azurePath =
+            if (uriPath.startsWith("/")) uriPath.drop(1) else uriPath
+
+          s"/vsiaz/${targetURI.getUserInfo}/$azurePath"
+        case HDFS =>
+          s"/vsihdfs/${targetURI.toString}"
+        case FILE =>
+          if (isWindows)
+            targetURI.toString.split("file:/").tail.head
```
pomadchin: keep in mind that both `tail` and `head` operations are _unsafe_. It means if for some reason it won't be possible to extract tail / head it will throw an exception that can be pretty confusing.

jbouffard: That's a good point. I could probably figure out a more safe way of extracting the right info from the `uri`.

vlm/src/main/scala/geotrellis/contrib/vlm/gdal/VSIPath.scala
jbouffard: I think both would work. It might be worth it to make it a class if we'd like to do more with the paths.

vlm/src/main/scala/geotrellis/contrib/vlm/gdal/VSIPath.scala
```diff
@@ -0,0 +1,208 @@
+package geotrellis.contrib.vlm.gdal
+
+import scala.util.matching.Regex
+
+import java.io.File
+
+
+case class VSIPath(
+  path: String,
+  compressedFileDelimiter: String = "!"
```
echeipesh: When can it be something other than `"!"` ?

jbouffard: That's there to catch the cases where there's a `!` in one of the file names for some reason.

vlm/src/main/scala/geotrellis/contrib/vlm/gdal/VSIPath.scala
```diff
@@ -0,0 +1,208 @@
...
+  // Trying to read something locally on Windows matters
+  // because of how file paths on Windows are formatted.
+  // Therefore, we need to handle them differently.
+  private lazy val onLocalWindows: Boolean =
+    System.getProperty("os.name").toLowerCase == "win" && isLocal
+
+  lazy val scheme: Option[String] =
+    SCHEME_PATTERN.findFirstIn(path)
+
+  private lazy val schemeString: String =
```
echeipesh: These kind of trivial field should always default to being a `def` while `lazy val` is "cheap" holding a trivial field on heap is kind of expensive. Rule of thumb I use is to treat `lazy val` as optimization, avoid it unless there is a specific reason to get the caching behavior.

echeipesh: I see its used a lot though, so more of a case for `val`

echeipesh: @jbouffard I'm with you, for re-use `val` makes sense. Please modify my comment towards as reaction towards seeing lots of `lazy` modifies.

vlm/src/main/scala/geotrellis/contrib/vlm/gdal/GDAL.scala
```diff
@@ -88,8 +88,11 @@ object GDAL {
       case _ => throw new Exception(s"Could not find equivalent GDALResampleMethod for: $method")
     }
 
-  def open(path: String): Dataset = {
-    val ds = gdal.Open(path, gdalconstConstants.GA_ReadOnly)
+  def open(path: String): Dataset =
+    open(VSIPath(path))
```
echeipesh: This would fail on correct /vsi paths look like

```scala
  describe("getting VSI uri") {
    it("doesn't change vsi uri") {
      val path = "/vsizip//tmp/some/data/data.zip"
      VSIPath(path).vsiPath shouldBe path
    }
  }
```

```
[info] - doesn't change vsi uri *** FAILED *** (14 milliseconds)
[info]   "/vsizip//[vsizip//]tmp/some/data/data.z..." was not equal to "/vsizip//[]tmp/some/data/data.z..." (VSIPathSpec.scala:306)
```

echeipesh: I think the thing to do is to have 
- `GDAL.openPath` pass the string to GDAL unmodified
- `GDAL.openUri` performing the praising, assuming that what you get is a URI
- `GDAL.open` perform best effort check if the path is URI or path and dispatch accordingly.

I'd tend to leave `VSIPath` out-of signature for `GDAL.open` because I would not expect the user to have to know about this type, its mostly a complex parsing function. That also makes it easier to refactor and change it when bugs are found.

vlm/src/main/scala/geotrellis/contrib/vlm/gdal/Patterns.scala
```diff
@@ -0,0 +1,21 @@
+package geotrellis.contrib.vlm.gdal
+
+import scala.util.matching.Regex
+
+
+/*
+ * Contains the different Regexs needed to parse
+ * the paths given to GDAL to read.
+ */
+object Patterns {
```
echeipesh: As a general comment, not a request for change, I would say that this parsing would be much easier to think about and more durable if you were to use RegEx groups and tried to parse the whole URI as a whole unit in a single RegEx path.

jbouffard: I wasn't aware we could do that in Scala. I'll look and see if that would simplify things.

PR #44, opened 2018-11-06
RasterRef Renamed to RasterRegion

jbouffard:
This PR renames the `RasterRef` case class to `RasterRegion` in order to better describe what the object represents. In addition to `RasterRef`, other methods had their names changed as well to reflect the class name change.

### DIFF
--- a/vlm/src/main/scala/geotrellis/contrib/vlm/RasterSourceRegion.scala
+++ b/vlm/src/main/scala/geotrellis/contrib/vlm/RasterSourceRegion.scala
@@ -20,16 +20,19 @@ import geotrellis.vector.{Extent, ProjectedExtent}
 import geotrellis.raster._
 
 /** Reference to a pixel region in a [[RasterSource]] that may be read at a later time.
+  * @note It is required that the [[RasterSource]] intersects with the given [[GridBounds]].
   *
-  * @param source raster source that can be used to read this ref
+  * @param source raster source that can be used to read this region.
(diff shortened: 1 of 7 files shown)


vlm/src/main/scala/geotrellis/contrib/vlm/RasterSourceRegion.scala
```diff
@@ -20,16 +20,19 @@ import geotrellis.vector.{Extent, ProjectedExtent}
...
+  * @param source raster source that can be used to read this region.
   * @param bounds pixel bounds relative to the source, maybe not be fully contained by the source bounds.
   */
-case class RasterRef(
+case class RasterSourceRegion(
   source: RasterSource,
   bounds: GridBounds
 ) extends CellGrid with Serializable {
+  require(bounds.intersects(source.gridBounds), s"The given bounds: $bounds must intersect the given source: $source")
+
```
echeipesh: This restriction makes it impossible to read a square tile when it partially intersects a raster source perhaps on TileLayerSource?. So either there needs to be a special way to read a ref or all refs must support ability to buffer when read.

jbouffard: Why would it be impossible? If it partially intersects then the requirement should be met.

metasim:
@echeipesh  I'm ambivalent about the long term name except that `RasterRef` has fully permeated the jargon here. Rolls off the tongue. Thanks for asking tho!

Issue #45, opened 2018-11-13
Improper GDALWarpOptions target pixels alignment usage


pomadchin:
We don't use properly GDALWarp, and in particular `-tap` parameter (`align the coordinates of the extent of the output file to the values of the -tr, such that the aligned extent includes the minimum extent`) [link](https://www.gdal.org/gdalwarp.html).

An example of a [failing test](https://github.com/geotrellis/geotrellis-contrib/blob/master/vlm/src/test/scala/geotrellis/contrib/vlm/RasterSummarySpec.scala#L132-L135). At the moment test fails, but without alignment of the target pixels it will pass.

PR #54, opened 2018-12-03
Add a pixelAlignment flag

pomadchin:
`-tap just aligns the grid to the resolution`, so probably we don't need it always, as sometimes it can round extent incorrectly.

Fixes https://github.com/geotrellis/geotrellis-contrib/issues/45

### DIFF
--- a/vlm/src/main/scala/geotrellis/contrib/vlm/gdal/GDALBaseRasterSource.scala
+++ b/vlm/src/main/scala/geotrellis/contrib/vlm/gdal/GDALBaseRasterSource.scala
@@ -27,11 +27,35 @@ import org.gdal.gdal.{Dataset, gdal}
 import org.gdal.osr.SpatialReference
 
 trait GDALBaseRasterSource extends RasterSource {
-  // options from previous transformation steps
+  /** Aligns the coordinates of the extent of the output file to the values of the target resolution,
+    * such that the aligned extent includes the minimum extent.
+    *
+    * In terms of GeoTrellis it's very similar to [[GridExtent]].createAlignedGridExtent:
+    *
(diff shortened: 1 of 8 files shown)


vlm/src/main/scala/geotrellis/contrib/vlm/gdal/GDALResampleRasterSource.scala
```diff
@@ -29,7 +29,8 @@ case class GDALResampleRasterSource(
   uri: String,
   resampleGrid: ResampleGrid,
   method: ResampleMethod = NearestNeighbor,
-  baseWarpList: List[GDALWarpOptions] = Nil
+  baseWarpList: List[GDALWarpOptions] = Nil,
+  alignTargetPixels: Boolean = true
```
echeipesh: In order to not perpetuate mistakes of our fathers this `GDALResampleRasterSource` needs a ScalaDoc that explains this option in particular.

vlm/src/main/scala/geotrellis/contrib/vlm/gdal/GDALBaseRasterSource.scala
```diff
@@ -27,11 +27,13 @@ import org.gdal.gdal.{Dataset, gdal}
 import org.gdal.osr.SpatialReference
 
 trait GDALBaseRasterSource extends RasterSource {
-  // options from previous transformation steps
+  /** -tap aligns the grid to the resolution, if the resolution is 5m, the cells are aligned to round multiples of 5. */
```
echeipesh: What does `aligned` mean? What is their reference for the top left corner of the top-left pixel?

PR #60, opened 2018-12-06
Overview strategy selection

pomadchin:
This PR exposes overviews resolutions and makes overview selection strategy configurable.

### DIFF
--- a/vlm/src/main/scala/geotrellis/contrib/vlm/LayoutTileSource.scala
+++ b/vlm/src/main/scala/geotrellis/contrib/vlm/LayoutTileSource.scala
@@ -83,7 +83,7 @@ class LayoutTileSource(val source: RasterSource, val layout: LayoutDefinition) e
     * If each tile area intersects source partially the non-intersecting pixels will be filled with NODATA.
     * If tile area does not intersect source it will be excluded from result iterator.
     */
-  def readAll(keys: Iterator[SpatialKey], bands: Seq[Int]) =
+  def readAll(keys: Iterator[SpatialKey], bands: Seq[Int]): Iterator[(SpatialKey, MultibandTile)] =
     for {
       key <- keys
       col = key.col.toLong
@@ -119,7 +119,7 @@ class LayoutTileSource(val source: RasterSource, val layout: LayoutDefinition) e
     readAll(keys.toIterator)
 
   /** Set of keys that can be read from this tile source */
-  def keys(): Set[SpatialKey] = {
+  def keys: Set[SpatialKey] = {
     layout.extent.intersection(source.extent) match {
       case Some(intersection) =>
         layout.mapTransform.keysForGeometry(intersection.toPolygon)
@@ -130,7 +130,7 @@ class LayoutTileSource(val source: RasterSource, val layout: LayoutDefinition) e
 
   /** All intersecting RasterRegions with their respective keys */
   def keyedRasterRegions(): Iterator[(SpatialKey, RasterRegion)] =
-    keys().toIterator.map(key => (key, rasterRegionForKey(key)))
+    keys.toIterator.map(key => (key, rasterRegionForKey(key)))
 
   def close = source.close()
 }
--- a/vlm/src/main/scala/geotrellis/contrib/vlm/RasterSource.scala
+++ b/vlm/src/main/scala/geotrellis/contrib/vlm/RasterSource.scala
@@ -49,39 +49,43 @@ trait RasterSource extends CellGrid with AutoCloseable with Serializable {
     def cellType: CellType
     def cellSize: CellSize = rasterExtent.cellSize
     def rasterExtent: RasterExtent
-    def overviewsRasterExtents: List[RasterExtent]
+    def cellSizes: List[CellSize]
+    def overviewCellSizes: List[CellSize] = cellSizes match {
+      case Nil | _ :: Nil => Nil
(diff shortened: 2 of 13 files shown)


vlm/src/main/scala/geotrellis/contrib/vlm/LayoutTileSource.scala
```diff
@@ -119,7 +119,7 @@ class LayoutTileSource(val source: RasterSource, val layout: LayoutDefinition) e
     readAll(keys.toIterator)
 
   /** Set of keys that can be read from this tile source */
-  def keys(): Set[SpatialKey] = {
```
echeipesh: Not that I'm against it, but why the switch?
Should this be a default style we choose for 0-argument methods ?

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSource.scala
```diff
@@ -82,29 +84,29 @@ trait RasterSource extends CellGrid with AutoCloseable with Serializable {
...
+      resample(TargetGrid(grid), method, strategy)
 
     /** Sampling grid and resolution is defined by given [[RasterExtent]] region.
       * The extent of the result is also taken from given [[RasterExtent]],
       *   this region may be larger or smaller than the footprint of the data
       * @group resample
       */
-    def resampleToRegion(region: RasterExtent, method: ResampleMethod = NearestNeighbor): RasterSource =
-      resample(TargetRegion(region), method)
+    def resampleToRegion(region: RasterExtent, method: ResampleMethod = NearestNeighbor, strategy: OverviewStrategy = AutoHigherResolution): RasterSource =
```
echeipesh: Sounds like we should move it out of `Geotrellis.raster.io.geotiff` package at some point if we're going to use it like this :/

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSource.scala
```diff
@@ -46,8 +47,9 @@ trait RasterSource extends CellGrid with AutoCloseable with Serializable {
     def crs: CRS
     def bandCount: Int
     def cellType: CellType
-    def cellSize = rasterExtent.cellSize
+    def cellSize: CellSize = rasterExtent.cellSize
     def rasterExtent: RasterExtent
+    def overviewsRasterExtents: List[RasterExtent]
```
echeipesh: `overviewsRasterExtents` seems too specific. What about GeoTrellis indexed layers?

The other problem is that they can easily blow out the `Int` cell count of `RasterExtent`.
The two options remaining are:
- `CellSize`: assume that extents are always the same for levels of detail
- `List[GridExtent]`: wide enough and convertible to `RasterExtent`

vlm/src/main/scala/geotrellis/contrib/vlm/gdal/GDALReprojectRasterSource.scala
```diff
@@ -67,8 +68,8 @@ case class GDALReprojectRasterSource(
   }
 
   override def reproject(targetCRS: CRS, options: Reproject.Options): RasterSource =
-    GDALReprojectRasterSource(uri, targetCRS, options, warpList, alignTargetPixels)
+    GDALReprojectRasterSource(uri, targetCRS, options, alignTargetPixels, warpList)
```
echeipesh: Missed this in the last PR but shouldn't `alignTargetPixels` be part of the `options` ?
Does it make things easier to break it out ?

vlm/src/main/scala/geotrellis/contrib/vlm/geotiff/GeoTiffReprojectRasterSource.scala
```diff
@@ -97,11 +99,8 @@ class GeoTiffReprojectRasterSource(
...
   def reproject(targetCRS: CRS, options: Reproject.Options): RasterSource =
-    new GeoTiffReprojectRasterSource(uri, targetCRS, options)
+    GeoTiffReprojectRasterSource(uri, targetCRS, options)
 
-  def resample(resampleGrid: ResampleGrid, method: ResampleMethod): RasterSource =
-    new GeoTiffReprojectRasterSource(uri, crs, options.copy(
-      method = method,
-      targetRasterExtent = Some(resampleGrid(self.rasterExtent))
-    ))
+  def resample(resampleGrid: ResampleGrid, method: ResampleMethod, strategy: OverviewStrategy): RasterSource =
```
echeipesh: `strategy` not used here.

vlm/src/main/scala/geotrellis/contrib/vlm/gdal/GDALReprojectRasterSource.scala
pomadchin: It can, but there appears a question: does it make any sense? in some operations (resample, reprojection) we calculate target cellSize basing on the the raster params / smth else, and these options should override / partially override these calculated values?

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSource.scala
pomadchin: Hm do you want me to move it in 3.0 release?

pomadchin:
For some reason we're doing things here: https://github.com/geotrellis/geotrellis-contrib/pull/64 at the moment.

PR #64, opened 2018-12-12
RasterSource resolutions and overview strategy

echeipesh:
Continuation of: https://github.com/geotrellis/geotrellis-contrib/pull/60

### DIFF
--- a/vlm/src/main/scala/geotrellis/contrib/vlm/RasterSource.scala
+++ b/vlm/src/main/scala/geotrellis/contrib/vlm/RasterSource.scala
@@ -49,39 +49,43 @@ trait RasterSource extends CellGrid with AutoCloseable with Serializable {
     def cellType: CellType
     def cellSize: CellSize = rasterExtent.cellSize
     def rasterExtent: RasterExtent
-    def overviewsRasterExtents: List[RasterExtent]
+    def cellSizes: List[CellSize]
+    def overviewCellSizes: List[CellSize] = cellSizes match {
+      case Nil | _ :: Nil => Nil
+      case _ :: tail      => tail
+    }
     def extent: Extent = rasterExtent.extent
     def cols: Int = rasterExtent.cols
     def rows: Int = rasterExtent.rows
     def bounds: GridBounds = GridBounds(0, 0, cols - 1, rows - 1)
 
-    /** Reproject to different CRS with explicit sampling options.
+    /** Reproject to different CRS with explicit sampling reprojectOptions.
       * @see [[geotrellis.raster.reproject.Reproject]]
       * @group reproject
       */
-    def reproject(crs: CRS, options: Reproject.Options): RasterSource
+    def reproject(crs: CRS, reprojectOptions: Reproject.Options, strategy: OverviewStrategy): RasterSource
 
     /** Sampling grid is defined over the footprint of the data at default resolution
       * @group reproject
       */
-    def reproject(crs: CRS, method: ResampleMethod = NearestNeighbor): RasterSource =
-        reproject(crs, Reproject.Options(method = method))
+    def reproject(crs: CRS, method: ResampleMethod = NearestNeighbor, strategy: OverviewStrategy = AutoHigherResolution): RasterSource =
+        reproject(crs, Reproject.Options(method = method), strategy)
 
     /** Sampling grid and resolution is defined by given [[GridExtent]].
       * Resulting extent is the extent of the minimum enclosing pixel region
       *   of the data footprint in the target grid.
       * @group reproject a
       */
-    def reprojectToGrid(crs: CRS, grid: GridExtent, method: ResampleMethod = NearestNeighbor): RasterSource =
-        reproject(crs, Reproject.Options(method = method, parentGridExtent = Some(grid)))
+    def reprojectToGrid(crs: CRS, grid: GridExtent, method: ResampleMethod = NearestNeighbor, strategy: OverviewStrategy = AutoHigherResolution): RasterSource =
+        reproject(crs, Reproject.Options(method = method, parentGridExtent = Some(grid)), strategy)
 
     /** Sampling grid and resolution is defined by given [[RasterExtent]] region.
       * The extent of the result is also taken from given [[RasterExtent]],
       *   this region may be larger or smaller than the footprint of the data
       * @group reproject
       */
-    def reprojectToRegion(crs: CRS, region: RasterExtent, method: ResampleMethod = NearestNeighbor): RasterSource =
-        reproject(crs, Reproject.Options(method = method, targetRasterExtent = Some(region)))
+    def reprojectToRegion(crs: CRS, region: RasterExtent, method: ResampleMethod = NearestNeighbor, strategy: OverviewStrategy = AutoHigherResolution): RasterSource =
+        reproject(crs, Reproject.Options(method = method, targetRasterExtent = Some(region)), strategy)
 
 
     def resample(resampleGrid: ResampleGrid, method: ResampleMethod, strategy: OverviewStrategy): RasterSource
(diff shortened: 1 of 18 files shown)


vlm/src/main/scala/geotrellis/contrib/vlm/RasterSource.scala
```diff
@@ -41,144 +42,176 @@ import geotrellis.util.GetComponent
...
-     * Applies the given [[LayoutDefinition]] to the source data producing a [[LayoutTileSource]].
-     * In order to fit to the given layout, the source data is resampled to match the Extent
-     * and CellSize of the layout.
-     *
-     */
-    def tileToLayout(layout: LayoutDefinition, resampleMethod: ResampleMethod = NearestNeighbor): LayoutTileSource =
-      LayoutTileSource(resampleToGrid(layout, resampleMethod), layout)
-
-    def close = { }
+  def resampleMethod: Option[ResampleMethod]
```
echeipesh: Sorry, this got auto-fixed to consistent indentation and makes this diff very ... not useful

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSource.scala
```diff
@@ -41,144 +42,176 @@ import geotrellis.util.GetComponent
...
+    * <li> For reprojected [[RasterSource]] these resolutions represent an estimate where
+    *      each cell in target CRS has ''approximately'' the same geographic coverage as a cell in the source CRS.
+    *
+    * When reading raster data the underlying implementation will have to sample from one of these resolutions.
+    * It is possible that a read request for a small bounding box will results in significant IO request when the target
+    * cell size is much larger than closest available resolution.
+    *
+    * __Note__: It is expected but not guaranteed that the extent each [[RasterExtent]] in this list will be the same.
+    */
+  def resolutions: List[RasterExtent]
```
echeipesh: - added comment here
Backtracking on my previous though. The `extent` here is not "useful" but `RasterExtent` nicely packs the pixel dimensions of the overviews, which we can't really get otherwise. That makes it the closest useful thing :/

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSource.scala
```diff
@@ -41,144 +42,176 @@ import geotrellis.util.GetComponent
...
+    * @see [[geotrellis.raster.reproject.Reproject]]
+    * @group reproject
+    */
+  def reproject(crs: CRS, reprojectOptions: Reproject.Options, strategy: OverviewStrategy): RasterSource
+
+  /** Sampling grid is defined over the footprint of the data at default resolution
+    *
+    * When using this method the cell size in target CRS will be estimated such that
+    * each cell in target CRS has ''approximately'' the same geographic coverage as a cell in the source CRS.
+    *
```
echeipesh: added this clarification

vlm/src/main/scala/geotrellis/contrib/vlm/geotiff/GeoTiffReprojectRasterSource.scala
```diff
@@ -22,33 +22,39 @@ import geotrellis.raster._
...
+  override lazy val rasterExtent: RasterExtent = reprojectOptions.targetRasterExtent match {
     case Some(targetRasterExtent) =>
       targetRasterExtent
     case None =>
-      ReprojectRasterExtent(baseRasterExtent, transform, options)
+      ReprojectRasterExtent(baseRasterExtent, transform, reprojectOptions)
   }
+
+  lazy val resolutions: List[RasterExtent] =
+    rasterExtent :: tiff.overviews.map(ovr => ReprojectRasterExtent(ovr.rasterExtent, transform))
```
echeipesh: Note that I dropped the `reprojectOptions` usage here. Basically they may have either target cell size or target raster extent, which would make all of these be the same value. What we're after here is the "default" behavior of trying to estimate.

echeipesh:
What I noticed is that `GeoTiffResampledRasterSource.resolutions` and `GDALRasterSource.resolutions` do NOT share the same behavior when resampling.

`GeoTiffRasterSource.resolutions` will of course re-use the resolutions of unmodified file, which makes sense in the way we defined it.

`GDALRasterSource.resolutions` will report on overviews that are created by GDAL, which will be automatically adjusted by the same ratio as the resample operation ex:

```
Resampled:
List(
GridExtent(Extent(630000.0, 215000.0, 645000.0, 228500.0),10.526315789473685,10.530421216848675), 
GridExtent(Extent(630000.0, 215000.0, 645000.0, 228500.0),21.037868162692845,21.06084243369735), 
GridExtent(Extent(630000.0, 215000.0, 645000.0, 228500.0),42.13483146067416,42.05607476635514), 
GridExtent(Extent(630000.0, 215000.0, 645000.0, 228500.0),83.79888268156425,83.85093167701864)) 

Source:
List(
GridExtent(Extent(630000.0, 215000.0, 645000.0, 228500.0),10.0,10.0), 
GridExtent(Extent(630000.0, 215000.0, 645000.0, 228500.0),20.0,20.0), 
GridExtent(Extent(630000.0, 215000.0, 645000.0, 228500.0),40.0,39.94082840236686), 
GridExtent(Extent(630000.0, 215000.0, 645000.0, 228500.0),79.7872340425532,79.88165680473372)) 
```

pomadchin:
@echeipesh yea, they should behave different; it’s an expected thing; btw you could just push into my branch, GitHub gives you permissions by default :D

echeipesh:
I wasn't able to : /

echeipesh:
Added comment to clarify difference in behavior. As is I'm not totally sure what the correct choice here is. This may be pointing to the fact that we may ultimately want to manage overview selection when resampling and reproduction manually. That goes against "let GDAL do its thing" heuristic that is the safe bet. I guess let's wait and see what ends up useful?

I think in reality more consistent use of `overviewStrategy` is the actual meat of this PR.

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSource.scala
pomadchin: Is there smth better rather than resolutions?

echeipesh:
Thats the implementation details leaking through.
GeoTiff version:
selects overview it will sample, reads it, and resamples to target cell size.

GDAL version:
creates a band which adjusts all overviews and reads from them as if they are of target cell size, but adjusts the metadata.

They both do ... about the same thing. Although I'm unclear what would happen if you resampled by 10x ... would GDAL adjust all overviews by 10x, is that even useful?

pomadchin:
@echeipesh yea, so I think it just rebuilds all metadata pyramid basing on the changed overview. So it's not just by 10x but every next overview = previous / 2, and one of these previous can be resampled.

pomadchin:
I double checked the GDAL behaviour (by modifying our tests). For instance we have the following set of overviews: 10, 20, 40, 80

if we'll resample up to 50: gdal will pick up 40, upsample it to 50 and will build the next overview as 100.

```
    // read in the whole file and resample the pixels in memory
    val expected: Raster[MultibandTile] =
      GeoTiffReader
        .readMultiband(url, streaming = false)
        .raster
        .resample((source.cols * 0.2).toInt , (source.rows * 0.2).toInt, NearestNeighbor)
    // resample to 0.9 so we RasterSource picks the base layer and not an overview

    val resampledSource =
      source.resample(TargetRegion(expected.rasterExtent), NearestNeighbor, AutoHigherResolution)

    resampledSource should    have (dimensions (expected.tile.dimensions))

    info(s"Source CellSize: ${source.cellSize}")
    info(s"Target CellSize: ${resampledSource.cellSize}")

    println(resampledSource.resolutions.map(_.cellSize)) //> List(CellSize(50.0,50.0), CellSize(100.0,100.0))
```

echeipesh:
The original rationale was to allow the user to know at which resolutions its "safe" to sample without triggering huge amounts of IO, which is basically originally available overviews. In that way exposing the rebuilt overviews in the way you describe isn't actually useful and in fact hopefully they will never be used, instead GDAL will always need to reach to one of the original overviews.

PR #66, opened 2018-12-18
Add test catalog with tests

pcaisse:
# Overview

Create a test catalog from sample GeoTIFF (`aspect-tiled.tif`).

# Notes

`TestCatalog.scala` is based off of https://github.com/geotrellis/geotrellis-landsat-tutorial/blob/master/src/main/scala/tutorial/IngestImage.scala

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,5 @@ nohup.out
 derby.log
 metastore_db/
 *.log
+
+vlm/src/test/resources/data/
--- a/vlm/src/test/scala/geotrellis/contrib/vlm/CatalogTestEnvironment.scala
+++ b/vlm/src/test/scala/geotrellis/contrib/vlm/CatalogTestEnvironment.scala
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2018 Azavea
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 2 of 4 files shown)


pomadchin:
Hey @pcaisse that's cool, but why is it here? We already have test coverage for all our backends in geotrellis-core: 
* [hadoop-example](https://github.com/locationtech/geotrellis/tree/master/spark/src/test/scala/geotrellis/spark/io/hadoop) 
* [insights of the persistence spec](https://github.com/locationtech/geotrellis/blob/master/spark-testkit/src/main/scala/geotrellis/spark/testkit/io/PersistenceSpec.scala)
* [Ingest example without persistence though](https://github.com/locationtech/geotrellis/blob/master/spark/src/test/scala/geotrellis/spark/ingest/IngestSpec.scala) 

VLM is a new API and a new concept of how we plan to work with rasters: [example link](https://github.com/geotrellis/geotrellis-contrib/blob/master/vlm/src/test/scala/geotrellis/contrib/vlm/RasterSummarySpec.scala#L191) And here we plan to implement and cover with tests only this new functionality.

echeipesh:
@pomadchin This is setup for writing GeoTrellisRasterSource / AvroRasterSource. The idea is to setup some test data that should be exactly the same as aspect-tiled.tiff but actually tiled as GeoTrellis layers. It will allow us to test that GeoTrellisRasterSource produces exactly the same pixels as GeoTiffRasterSource when reading tiled data.

pomadchin:
LGTM than!

PR #75, opened 2018-12-27
Fix PaddedTile foreach and map visitors

echeipesh:
- `foreach` and `map` visitors were getting col/row values relative to chunk rather than padded tile itself
- Imposed row-major traversal order in `foreach(z)` and `foreachDouble(x)` methods

These were discovered via a bug that shifted the data region to upper left corner when a tile was being rendered.


moradology:
+1 once it passes

PR #76, opened 2018-12-28
Update geotrellis-gdal version

pomadchin:


pomadchin:
The problems still present.

PR #85, opened 2019-01-08
Add GeotrellisRasterSource

pcaisse:
## Overview

Read in a GT catalog by layer ID.

## Notes

Some tests borrowed from `GeoTiffRasterSource`. Resampling and reprojection to come in subsequent PRs.

### DIFF
--- a/vlm/src/main/scala/geotrellis/contrib/vlm/GeotrellisRasterSource.scala
+++ b/vlm/src/main/scala/geotrellis/contrib/vlm/GeotrellisRasterSource.scala
@@ -1,0 +1,116 @@
+/*
+ * Copyright 2018 Azavea
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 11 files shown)


vlm/src/main/scala/geotrellis/contrib/vlm/GeotrellisRasterSource.scala
```diff
@@ -0,0 +1,116 @@
...
+
+/**
+  * Note: GeoTrellis AttributeStore does not store the band count for the layers by default,
+  *       thus they need to be provided from application configuration.
+  *
+  * @param uri geotrellis catalog uri
+  * @param layerId source layer from above catalog
+  * @param bandCount number of bands for each tile in above layer
+  */
+case class GeotrellisRasterSource(uri: String, layerId: LayerId, bandCount: Int = 1) extends RasterSource {
```
echeipesh: For consistency this should move into its own package.
I propose `goetrellis.contrib.vlm.avro`

pcaisse: Sure. Should I also move `TestCatalog.scala` into `goetrellis.contrib.vlm.avro` then?

echeipesh: Thats a good point, it should go there.

echeipesh:
Looks like the test is failing with `pure virtual method called` error at cleanup time. This looks like its related to GDAL cleanup and and not anything going on in this PR.

PR #100, opened 2019-01-17
Get rid of multithreaded issues caused by dependent VRTs

pomadchin:


pomadchin:
Merging, will test it next sprint with RF

PR #103, opened 2019-01-21
Add GeotrellisResampleRasterSource

echeipesh:
Adds implementation of RasterSource that selects closest zoom level and resamples to match desired resolution and pixel offset.


pcaisse:
Closed in favor of https://github.com/geotrellis/geotrellis-contrib/pull/105

PR #110, opened 2019-01-31
tiledLayoutRDD Method for RasterSourceRDD

jbouffard:
This PR adds the `tiledLayerRDD` method to `RasterSourceRDD`. The reason for this addition is because it will allow users to create `MultibandTileLayer`s from a larger amount of `RasterSource`s in a more convenient way.

### DIFF
--- a/vlm/src/main/scala/geotrellis/contrib/vlm/RasterSourceRDD.scala
+++ b/vlm/src/main/scala/geotrellis/contrib/vlm/RasterSourceRDD.scala
@@ -118,7 +118,8 @@ object RasterSourceRDD {
   def tiledLayerRDD(
     sources: RDD[RasterSource],
     layout: LayoutDefinition,
-    resampleMethod: ResampleMethod = NearestNeighbor
+    resampleMethod: ResampleMethod = NearestNeighbor,
+    partitioner: Option[Partitioner] = None
   )(implicit sc: SparkContext): MultibandTileLayerRDD[SpatialKey] = {
     val summary = RasterSummary.fromRDD(sources)
     val layerMetadata = summary.toTileLayerMetadata(layout, 0)._1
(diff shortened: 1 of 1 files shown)


vlm/src/main/scala/geotrellis/contrib/vlm/RasterSourceRDD.scala
```diff
@@ -114,6 +114,39 @@ object RasterSourceRDD {
...
+    val layerKeyBounds = KeyBounds(mapTransform(combinedExtents))
+
+    val layerMetadata =
+      TileLayerMetadata[SpatialKey](cellType, layout, combinedExtents, crs, layerKeyBounds)
+
+    val keyedRDD: RDD[(SpatialKey, MultibandTile)] =
+      sources.flatMap { case source =>
+        val layoutTileSource = LayoutTileSource(source, layout)
+
+        layoutTileSource.readAll(layoutTileSource.keys.toIterator)
```
pomadchin: Look here to find an example of `RDD[RasterSource]` conversion into `TileLayerRDD`:

https://github.com/geotrellis/geotrellis-contrib/blob/master/vlm/src/test/scala/geotrellis/contrib/vlm/RasterSummarySpec.scala#L96-L103

In this code you're loading the entire `RasterSource` into a single partition. Will definitely be a bottleneck.

jbouffard: Oh, I completely missed that. Thank you for pointing that out.

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSourceRDD.scala
```diff
@@ -114,6 +115,28 @@ object RasterSourceRDD {
...
+
+    val tiledLayoutSourceRDD =
+      sources.map { _.tileToLayout(layout, resampleMethod) }
+
+    val rasterRegionRDD: RDD[(SpatialKey, RasterRegion)] =
+      tiledLayoutSourceRDD.flatMap { _.keyedRasterRegions() }
+
+    val tiledRDD: RDD[(SpatialKey, MultibandTile)] =
+      rasterRegionRDD
+        .groupByKey(SpatialPartitioner(summary.estimatePartitionsNumber))
```
echeipesh: I would suggest not hard-coding the partitioner but having a param `partitioner: Option[Partitioner] = None` that can pick a default result. if it's set to None.

PR #111, opened 2019-01-31
readFromRDD Method and Small Fix to the read Method

jbouffard:
This PR adds the `readFromRDD` method to `RasterSourceRDD` that will create a `MultibandTileLayer` from an `RDD[ReadingSource]`. In addition to this new method, a small fix has been done in the `read` method that improves the partitioning of that method.

### DIFF
--- a/vlm/src/main/scala/geotrellis/contrib/vlm/RasterSourceRDD.scala
+++ b/vlm/src/main/scala/geotrellis/contrib/vlm/RasterSourceRDD.scala
@@ -31,12 +31,18 @@ import scala.collection.mutable.ArrayBuilder
 import scala.reflect.ClassTag
 
 object RasterSourceRDD {
-  final val PARTITION_BYTES: Long = 128l * 1024 * 1024
+  final val DEFAULT_PARTITION_BYTES: Long = 128l * 1024 * 1024
+
+  def read(
+    readingSources: Seq[ReadingSource],
+    layout: LayoutDefinition
+  )(implicit sc: SparkContext): MultibandTileLayerRDD[SpatialKey] =
+    read(readingSources, layout, DEFAULT_PARTITION_BYTES)
 
   def read(
     readingSources: Seq[ReadingSource],
     layout: LayoutDefinition,
-    partitionBytes: Long = PARTITION_BYTES
+    partitionBytes: Long
   )(implicit sc: SparkContext): MultibandTileLayerRDD[SpatialKey] = {
     val cellTypes = readingSources.map { _.source.cellType }.toSet
     require(cellTypes.size == 1, s"All RasterSources must have the same CellType, but multiple ones were found: $cellTypes")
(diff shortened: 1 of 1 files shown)


vlm/src/main/scala/geotrellis/contrib/vlm/RasterSourceRDD.scala
```diff
@@ -65,7 +65,7 @@ object RasterSourceRDD {
     val targetIndexes: Seq[Int] = 0 to maxIndex
 
     val sourcesRDD: RDD[(SpatialKey, (Int, Option[MultibandTile]))] =
-      sc.parallelize(readingSources).flatMap { source =>
+      sc.parallelize(readingSources, readingSources.size).flatMap { source =>
```
pomadchin: 👍

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSourceRDD.scala
```diff
@@ -115,6 +115,61 @@ object RasterSourceRDD {
     ContextRDD(result, layerMetadata)
   }
 
+  def readFromRDD(
```
pomadchin: a bit a confusing name(?)

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSourceRDD.scala
```diff
@@ -115,6 +115,61 @@ object RasterSourceRDD {
...
+          }
+
+        key -> MultibandTile(tiles)
+      }
+    }, preservesPartitioning = true)
+
+    ContextRDD(result, layerMetadata)
+  }
+
   def tiledLayerRDD(
```
pomadchin: Should not it be a `tileLayerRDD`?

jbouffard: I was thinking `read` since it's dealing with `ReadingSource`s. Regardless, though, the reason I didn't do that, though, is because of:

```
[error] /home/jake/Documents/projects/geotrellis-contrib/vlm/src/main/scala/geotrellis/contrib/vlm/RasterSourceRDD.scala:33:8: in object RasterSourceRDD, multiple overloaded alternatives of method read define default arguments.
[error] object RasterSourceRDD {
```

So I'm going to have to figure out a way to refactor those methods so that they can have the same name.

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSourceRDD.scala
jbouffard: Yeah, it's not the best name.

jbouffard: I was thinking `read` since it's dealing with `ReadingSource`s. Regardless, though, the reason I didn't do that, though, is because of:

```
[error] /home/jake/Documents/projects/geotrellis-contrib/vlm/src/main/scala/geotrellis/contrib/vlm/RasterSourceRDD.scala:33:8: in object RasterSourceRDD, multiple overloaded alternatives of method read define default arguments.
[error] object RasterSourceRDD {
```

So I'm going to have to figure out a way to refactor those methods so that they can have the same name.

PR #115, opened 2019-02-06
Convert Method

jbouffard:
This PR adds the `convert` method to `RasterSource`. This will allow users to convert the `CellType`s of their sources.


echeipesh:
This approach fails for `GeoTiffRasterSource` and `GeoTrellisRasterSource`:

```scala
scala> val rs = GeoTiffRasterSource("/Users/eugene/B1_R299_wse.tif")
rs: geotrellis.contrib.vlm.geotiff.GeoTiffRasterSource = GeoTiffRasterSource(/Users/eugene/B1_R299_wse.tif)

scala> rs.cellType
res10: geotrellis.raster.CellType = float32ud-9999.0

scala> rs.crs
res11: geotrellis.proj4.CRS = aea-CS

scala> rs.reproject(LatLng)
res12: geotrellis.contrib.vlm.RasterSource = GeoTiffReprojectRasterSource(/Users/eugene/B1_R299_wse.tif,LatLng,Options(NearestNeighbor,0.125,None,None,None),AutoHigherResolution)

scala> res12.crs
res13: geotrellis.proj4.CRS = LatLng

scala> res12.cellType
res14: geotrellis.raster.CellType = float32ud-9999.0

scala> res12.convert(IntCellType)
res15: geotrellis.contrib.vlm.RasterSource = GeoTiffConvertedRasterSource(/Users/eugene/B1_R299_wse.tif,int32raw,AutoHigherResolution)

scala> res15.crs
res16: geotrellis.proj4.CRS = aea-CS
//  .convert lost information about the reprojection that's supposed to be happening.

scala> res15.cellType
res17: geotrellis.raster.CellType = int32raw
// this is as expected

// lets try again to retroject from converted RasterSource
scala> res15.reproject(LatLng)
res19: geotrellis.contrib.vlm.RasterSource = GeoTiffReprojectRasterSource(/Users/eugene/B1_R299_wse.tif,LatLng,Options(NearestNeighbor,0.125,None,None,None),AutoHigherResolution)

scala> res19.cellType
res20: geotrellis.raster.CellType = float32ud-9999.0
// now we lost information about conversion that we wanted

scala> res19.crs
res21: geotrellis.proj4.CRS = LatLng
```

For GDAL case this actually works out:
```scala
scala> val rs = GDALRasterSource("/Users/eugene/B1_R299_wse.tif")
rs: geotrellis.contrib.vlm.gdal.GDALRasterSource = GDALRasterSource(/Users/eugene/B1_R299_wse.tif,GDALWarpOptions(Some(VRT),None,None,None,true,None,None,None,None,List(),List(),Some(AutoHigherResolution),List(),false,None,false,false,false,None,List(),None,None,false,false,false,None,false,false,List(),None,None,None,None,None,false,false,false,None,false,List(),List(),List(),None))

scala> rs.cellType
res22: geotrellis.raster.CellType = float32ud-9999.0

scala> rs.crs
res23: geotrellis.proj4.CRS = aea-CS

scala> rs.reproject(LatLng)
res24: geotrellis.contrib.vlm.RasterSource = GDALReprojectRasterSource(/Users/eugene/B1_R299_wse.tif,LatLng,Options(NearestNeighbor,0.125,None,None,None),AutoHigherResolution,GDALWarpOptions(Some(VRT),None,None,None,true,None,None,None,None,List(),List(),Some(AutoHigherResolution),List(),false,None,false,false,false,None,List(),None,None,false,false,false,None,false,false,List(),None,None,None,None,None,false,false,false,None,false,List(),List(),List(),None),List())

scala> res24.crs
res25: geotrellis.proj4.CRS = longlat-CS

scala> res24.cellType
res26: geotrellis.raster.CellType = float32ud-9999.0

scala> res24.convert(IntCellType)
res28: geotrellis.contrib.vlm.RasterSource = GDALConvertedRasterSource(/Users/eugene/B1_R299_wse.tif,int32raw,AutoHigherResolution,GDALWarpOptions(Some(VRT),None,None,None,true,None,None,None,None,List(),List(),Some(AutoHigherResolution),List(),false,None,false,false,false,None,List(),None,None,false,false,false,None,false,false,List(),None,None,None,None,None,false,false,false,None,false,List(),List(),List(),None),List(GDALWarpOptions(Some(VRT),Some(NearestNeighbor),Some(0.125),None,true,None,Some(aea-CS),Some(longlat-CS),None,List(-9999.0),List(),Some(AutoHigherResolution),List(),false,None,false,false,false,None,List(),None,None,false,false,false,None,false,false,List(),None,None,None,None,None,false,false,false,None,false,List(),List(),List(),None)),[Lorg.gdal.gdal.Dataset;@16505149)

scala> res28.cellType
res29: geotrellis.raster.CellType = int32raw

scala> res28.crs
res30: geotrellis.proj4.CRS = longlat-CS
// got both LatLng and CellType
```

The reason it works in case of GDAL is because the actual work is done by the underlying GDAL `Dataset` object and preserving the `GDALWarpOptions` from chained calls preserves the information that allows it to create the VRT with right properties.

Going back to the `GeoTiffRasterSource` and `GeotrellisRasterSource` case through, what to do. On the face this creates combinatorial problem, now there could be:
`GeoTiffRasterSource`
`GeoTiffConvertedRasterSource`
`GeoTiffResampledRasterSource`
`GeoTiffResampledConvertedRasterSource`
`GeoTiffReprojectedRasterSource`
`GeoTiffReprojectedConvertedRasterSource`

oy ... supposing there are other lazy transformations that we may want (for instance adding `.interpretAs` in addition to `.convert`) it gets kind of crazy and unsustainable.

For this particular case all that's required is to call `.convert` or `.interpretAs` on the final raster after it has been read, there is no way to read in non-original CellType.

What this seems to imply to me is that there needs to be a way to carry `RasterViewOptions`, similar to `GDALWarpOptions` for non-GDAL sources. Then the job of each instance of `RasterSource` is to reconcile the difference between underlying data and the desired view of that data specified by the user. New subclasses can be created when the logic required for a specific set of `RasterViewOptions` is too complicated or mess to try to fit with other possibilities (which is where the motivation for separating Base, Resample and Reprojected reading came from originally).

Going off that there is additional API decision if `RasterViewOptions` should actually be exposed to the user as an object or if it should be an internal type that's used to capture the constructor parameters and then used reconcile the reading of pixels and metadata.

What do you think ?

jbouffard:
@echeipesh Ah, I have't gotten around to testing out chain calls yet. Thanks for pointing out this issue now :slightly_smiling_face: 

I think that `RasterViewOptions` sounds like a good solution to this issue for now. It also has the added benefit of making the underlying logic for all of the `RasterSource`s similar, which is nice.

In my opinion, both `RasterViewOptions` and `GDALWarpOptions` should be `private[vlm]` or something similar. Since those types describe how their respective `RasterSource` should be transformed and read in, we should protect them from users. Maybe instead we could have some kind record of the steps that will be performed before the data is read in?

Something like this maybe?
```scala

val source = GDALRasterSource("/tmp/some-data")

// check to see what the source will do when the data is read in
source.steps
// Because we haven't done anything yet, there are no steps that will be taken
List()

val reprojected = source.reproject(targetCRS)

// check to see what the source will do when the data is read in
reprojected.steps

// Now the reproject step will happen before the data is read in.
List("reproject")
```

Something like that might be useful for debugging. It'll also give the user the same information as `GDALWarpOptions` and `RasterViewOptions`, but they won't be able to mess with the actual transformation logic.

project/Dependencies.scala
```diff
@@ -18,7 +18,7 @@ import sbt._
 
 object Version {
   val geotrellis     = "2.2.0"
-  val geotrellisGdal = "0.17.19"
+  val geotrellisGdal = "0.17.15-11-gf8ddf5e-SNAPSHOT"
```
pomadchin: `0.18.0` is released

vlm/src/test/scala/geotrellis/contrib/vlm/gdal/GDALConvertedRasterSourceSpec.scala
```diff
@@ -0,0 +1,219 @@
...
+        val expected = source.read(targetExtent).get.mapTile { _.convert(ShortConstantNoDataCellType) }
+
+        assertRastersEqual(actual, expected, 1.0)
+      }
+
+      it("should convert to: ShortUserDefinedNoDataCellType(-1)") {
+        val actual = source.convert(ShortUserDefinedNoDataCellType(-1)).read(targetExtent).get
+        val expected = source.read(targetExtent).get.mapTile { _.convert(ShortUserDefinedNoDataCellType(-1)) }
+
+        assertRastersEqual(actual, expected, 1.0)
```
pomadchin: Can you add a note why there is such a huge threshold?

vlm/src/main/scala/geotrellis/contrib/vlm/geotiff/GeoTiffResampleRasterSource.scala
```diff
@@ -50,29 +50,31 @@ case class GeoTiffResampleRasterSource(
...
   }
 
+  private[geotiff] val options = parentOptions.copy(rasterExtent = Some(rasterExtent))
+
   @transient protected lazy val closestTiffOverview: GeoTiff[MultibandTile] =
     tiff.getClosestOverview(rasterExtent.cellSize, strategy)
 
-  def reproject(targetCRS: CRS, reprojectOptions: Reproject.Options, strategy: OverviewStrategy): GeoTiffReprojectRasterSource =
-    new GeoTiffReprojectRasterSource(uri, targetCRS, reprojectOptions, strategy) {
+  override def reproject(targetCRS: CRS, reprojectOptions: Reproject.Options, strategy: OverviewStrategy): GeoTiffReprojectRasterSource =
```
pomadchin: why have you added these verbose identifiers? O:

vlm/src/test/scala/geotrellis/contrib/vlm/avro/GeotrellisRasterSourceSpec.scala
```diff
@@ -34,8 +34,8 @@ class GeotrellisRasterSourceSpec extends FunSpec with RasterMatchers with Better
   val uriMultiband = s"file://${TestCatalog.multibandOutputPath}"
   val uriSingleband = s"file://${TestCatalog.singlebandOutputPath}"
   val layerId = LayerId("landsat", 0)
-  val sourceMultiband = GeotrellisRasterSource(uriMultiband, layerId)
-  val sourceSingleband = GeotrellisRasterSource(uriSingleband, layerId)
+  lazy val sourceMultiband = GeotrellisRasterSource(uriMultiband, layerId)
```
pomadchin: A pretty important thing here, it will be evaluated before `beforeAll` function call.

pomadchin:
@jbouffard I fixed tests, pay attention to the thing that `RasterSource` should be as lazy as it's possible. Can you double check your code?

vlm/src/main/scala/geotrellis/contrib/vlm/geotiff/GeoTiffResampleRasterSource.scala
jbouffard: Oh, I think I added those when I was first trying things out. I just forgot to clean them up. Thanks for catching that :+1:

vlm/src/test/scala/geotrellis/contrib/vlm/avro/GeotrellisRasterSourceSpec.scala
jbouffard: Ah, I wasn't aware of that :+1:

pomadchin: @jbouffard actually it's due to regression that was fixed here: https://github.com/geotrellis/geotrellis-contrib/pull/115/commits/d1cffd2393c350cb46de1eb28de8742ac5a9d863

jbouffard:
@pomadchin I'm not sure if it really makes sense to derive `GDALWarpOptions` from `stepCollection`. `stepCollection` is really just there for debugging, and it doesn't interact with the `options` really.

pomadchin:
@jbouffard if it's for debugging, do we need it at all?

jbouffard:
@pomadchin I think it'd be nice to have. Because it provides a record of the operations that are going to performed on the source and the order which they occur in. You used to be able to get that info from the `options`, but those are `private` now.

vlm/src/main/scala/geotrellis/contrib/vlm/geotiff/GeoTiffBaseRasterSource.scala
```diff
@@ -0,0 +1,91 @@
...
+    GeoTiffReprojectRasterSource(uri, targetCRS, reprojectOptions, strategy)
+
+  def resample(resampleGrid: ResampleGrid, method: ResampleMethod, strategy: OverviewStrategy): RasterSource =
+    GeoTiffResampleRasterSource(uri, resampleGrid, method, strategy)
+
+  def convert(cellType: CellType, strategy: OverviewStrategy): RasterSource =
+    GeoTiffConvertedRasterSource(uri, cellType, strategy)
+
+  def read(extent: Extent, bands: Seq[Int]): Option[Raster[MultibandTile]] = {
+    val bounds = rasterExtent.gridBoundsFor(extent, clamp = false)
```
echeipesh: `bounds` here is for pixels in the reprojected virtual raster as presented by `rasterExtent` but on next line its being used to read pixels from source raster pre-reprojection and then labelling them with extent of reprojected raster.

Because bounds all start with `(0,0)` and the reprojection approximately preserves the shape and size of the raster  this will actually return raster values most of the time. However it completely bypasses reprojection or resampling that may have been implied in the shape of `rasterExtent` and is fundamentally incorrect result.

vlm/src/main/scala/geotrellis/contrib/vlm/RasterViewOptions.scala
```diff
@@ -0,0 +1,13 @@
...
+import geotrellis.proj4.CRS
+import geotrellis.vector.Extent
+import geotrellis.raster.{RasterExtent, CellType, Raster, MultibandTile}
+
+
+private[vlm] case class RasterViewOptions(
+  crs: Option[CRS] = None,
+  cellType: Option[CellType] = None,
+  rasterExtent: Option[RasterExtent] = None,
+  readMethod:  Option[(Extent, Seq[Int]) => Option[Raster[MultibandTile]]] = None
```
echeipesh: I'm concerned about the effect on serialization attaching a this type of function to each `RasterSource` has. Its potentially easy to close over some non-serializable resource without realizing. In current implementation this would close over a reference to the source `RasterSource` and link to to converted `RasterSource`. Which isn't critical failure but is not great. Is there a another easier way to achieve the same thing here?

jbouffard:
Superseded by #126

PR #118, opened 2019-02-15
Various Fixes

jbouffard:
This PR adds two fixes: the first correcting the logic in `PaddedTile.foreachDouble`, and the second is fixing the key name in `reference.conf` to `allow-global-read`


vlm/src/main/scala/geotrellis/contrib/vlm/PaddedTile.scala
```diff
@@ -98,7 +98,7 @@ case class PaddedTile(chunk: Tile, colOffset: Int, rowOffset: Int, cols: Int, ro
 
   def foreachDouble(f: Double => Unit): Unit = {
     cfor(0)(_ < rows, _ + 1) { row =>
-      if (row < rowOffset || row > (rowOffset + rows - 1)) {
+      if (row < rowOffset || row > (rowOffset + chunk.rows) - 1) {
```
pomadchin: Have you added a unit test? Or investigation.spec just started to work?

vlm/src/main/resources/reference.conf
```diff
@@ -1,3 +1,3 @@
 vlm.geotiff.s3 {
-  allowGlobalRead: false
-}
\ No newline at end of file
+  allow-global-read: false
```
pomadchin: Rather than doing this, let's keep it as `allowGlobalRead: false` (as we have this style in `geotrellis-gdal`):

```scala
case class S3Options(allowGlobalRead: Boolean)

object S3Options extends PureConfigSettings with Serializable {
  val config = pureconfig.loadConfigOrThrow[S3Options]("vlm.geotiff.s3")
}
```

^ it's the current `Config.scala`. I think we need to rename it and move into `config` package.

pomadchin:
Ah, also I think it makes sense to squash everything into a single commit, since you have both `23b44eb` and `b845463` that should not be there

vlm/src/main/scala/geotrellis/contrib/vlm/PaddedTile.scala
jbouffard: Ah, `InvestigationSpec` passed, but then I removed it... I'll have to create a new unit test.

echeipesh:
> Ah, also I think it makes sense to squash everything into a single commit, since you have both `23b44eb` and `b845463` that should not be there

I've been doing that with "Squash and merge" through GitHub, it appears to me that having a single commit per PR makes things clearer in history than even well broken up commit history. Just sharing thoughts though.

vlm/src/main/resources/reference.conf
echeipesh: Doesn't pureconfig require snake case for the config variables?

pomadchin: @echeipesh it's configurable O:

pomadchin: Well, we can just come up with some common style and use it everywhere :)

jbouffard:
@echeipesh I agree that having more compacted commits per PR is something we should aim for, though, I disagree that 1 is the ideal number.

I think that each commit should encompass a complete, important change. For example, a commit that added a new feature would contain the feature itself, its unit tests, and all of the other work needed to make those two things run. If a version number also needed to be changed in the same PR, that version change would have its own commit. That way, it's easy to see what major changes occurred if/when that PR gets merged.

vlm/src/main/resources/reference.conf
pomadchin: @echeipesh also renaming this file would be very helpful as well :D dunno when it was merged.

echeipesh: We're making hyphen case as default.

pomadchin: Since it's used by the most libraries (for instance akka-http)

PR #120, opened 2019-02-21
Fix ambiguous GDALBaseReprojectRasterSource constructor

pomadchin:


vlm/src/main/scala/geotrellis/contrib/vlm/gdal/GDALReprojectRasterSource.scala
```diff
@@ -67,3 +66,8 @@ case class GDALReprojectRasterSource(
     )
   }
 }
+
+object GDALReprojectRasterSource {
+  def apply(uri: String, targetCRS: CRS): GDALReprojectRasterSource =
+    GDALRasterSource(uri).reproject(targetCRS).asInstanceOf[GDALReprojectRasterSource]
```
pomadchin: TODO: Think of a better way of doing it here.

vlm/src/main/scala/geotrellis/contrib/vlm/gdal/GDALReprojectRasterSource.scala
```diff
@@ -28,12 +27,12 @@ import org.gdal.osr.SpatialReference
 
 case class GDALReprojectRasterSource(
   uri: String,
-  targetCRS: CRS,
```
jbouffard: I think that it's a little weird that `targetCRS` isn't

PR #123, opened 2019-02-22
A partial and failed attempt at parameterizing over grid indexing primitive.

metasim:
Probably going to abandon this.

### DIFF
--- a/vlm/src/main/scala/geotrellis/contrib/vlm/LayoutTileSource.scala
+++ b/vlm/src/main/scala/geotrellis/contrib/vlm/LayoutTileSource.scala
@@ -16,9 +16,11 @@
 
 package geotrellis.contrib.vlm
 
-import geotrellis.raster._
-import geotrellis.spark.tiling._
+import geotrellis.contrib.vlm.model.GridBounds.LayerGridBounds
+import geotrellis.contrib.vlm.model.{GridBounds, GriddedExtent}
+import geotrellis.raster.{GridBounds => _, _}
 import geotrellis.spark.SpatialKey
+import geotrellis.spark.tiling._
 
 /** Reads tiles by key from a [[RasterSource]] as keyed by a [[LayoutDefinition]]
   * @note It is required that the [[RasterSource]] is pixel aligned with the [[LayoutDefinition]]
@@ -27,7 +29,7 @@ import geotrellis.spark.SpatialKey
   * @param layout definition of a tile grid over the pixel grid
   */
 class LayoutTileSource(val source: RasterSource, val layout: LayoutDefinition) extends AutoCloseable {
-  LayoutTileSource.requireGridAligned(source.rasterExtent, layout)
+  LayoutTileSource.requireGridAligned[Long](source.griddedExtent, GriddedExtent(layout.extent, layout.tileCols, layout.tileRows))
 
   def sourceColOffset: Long = ((source.extent.xmin - layout.extent.xmin) / layout.cellwidth).toLong
   def sourceRowOffset: Long = ((layout.extent.ymax - source.extent.ymax) / layout.cellheight).toLong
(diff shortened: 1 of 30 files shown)


echeipesh:
This could help a little: https://gist.github.com/echeipesh/e5f0150ae47d7c8438b8b8e95ced9f02

Where are the parts it fell apart?

metasim:
Fell apart in trying to use them in the RasterSource and all the subclasses, allowing read-by-bounds to handle Int and Long versions transparently.

echeipesh:
Picking this guy up, clearing up the compiler errors.

metasim:
@echeipesh Not convinced I was heading in the right direction, so please don't go too far with it unless you're convinced.

echeipesh:
@metasim  It looks about right to me. Mainly my opinion is that there probably needs to be a type level distinction between `GridBounds` and `PixelBounds`. Some of the type generality (use of `Integral`) may or may not be worth the weight, but shouldn't be hard to factor it out. Naming is hard, but refactor tools handle that. What were you doubts about, might be helpful to pay more attention now.

metasim:
@echeipesh Good to know you think there are worthwhile elements in the PR. I was lost in the forest by the end of it.

w.r.t. naming, `GridBounds` vs. `PixelBounds`, etc, some reflection on my thinking at the time:
1. Not having been a part of the original GT development, I regularly get confused maintaining a semantic difference between terms such as "pixel" vs "cell", "grid" vs "raster",  "extent" vs "bounds". In hindsight I see I was trying to simplify some of these distinctions, perhaps erroneously.  At the time I think I'd convinced myself there were two types of rectilinear lattices: layers (uses `Long` indexing) and rasters (uses `Int` indexing).
2. To anchor these two distinctions, to help convince myself they were correct, and to force a single definition (for the time being) I `sealed` the generic definition and required the specializations to defined all together, such as in [this example](https://github.com/geotrellis/geotrellis-contrib/pull/123/files#diff-4b47f7ac15b2a4d6d10908434749c91fR33). The companion object [constructor methods](https://github.com/geotrellis/geotrellis-contrib/pull/123/files#diff-4b47f7ac15b2a4d6d10908434749c91fR28) take care of selecting the concrete representation based on type parameter, so from a user standpoint you deal more with the generic representation. But when I tried to apply those distinctions to `RasterSource` I got stuck.

The heart of my concerns are:
1. There are two semantic contexts around dimensions that seem to collide, or at least require a better picture of the end state:
  a. The dimensions of the possible data-space represented by the `RasterSource`. This requires `Long` dimensions. 
  b. The dimensions of a readable sub-region of the `RasterSource` (e.g. `Tile`).
I think in the last version I committed, all the dimensions in `RasterSource` were `Long`, even when that's not correct (i.e. you can't remotely guarantee a read request can be fulfilled; `Int`s aren't even the right size, but kinda a compromise). I was trying to refine that so that `read(bounds: GridBounds[Long])` (or preferably `read(bounds: RasterBounds)`) were handled, but got me tied up in typeclass hell.
2. If I couldn't figure out a way to use the type system to ensure safety in handling layer vs. raster reads, then what's the point in all the type gymnastics that lead to this point. I "know" Scala can make this happen, but at what cognitive cost. The biggest complaint I get from my team when diving deep into GeoTrellis is tracking down (and implementing) the various typeclasses (e.g. why have `GetComponent[T, V]` when you could just have a `trait T` with `def component: V`). I felt like I was contributing to making the learning curve worse.

I still think the type generality over `Integral` is very valuable for classes such as [`GridBounds`](https://github.com/geotrellis/geotrellis-contrib/pull/123/files#diff-ceaee79437beed2737d4f960d3d58696R25) where there's a lot of algorithmic work that is essentially the same regardless of the numeric word size. If that could be better hidden behind concrete specializations then I'd feel like some of the effort was worth it. But I struggled to get there.

metasim:
PS: I like [Your version](https://gist.github.com/echeipesh/e5f0150ae47d7c8438b8b8e95ced9f02) of what I called `GridBounds` better as the use of spire cleans it up quite a bit.... so I think we should swap that out.

echeipesh:
@metasim As I continue to push on this I'm more convinced that this is just a staged refactor where the final state of things will be `RasterExtent[N]` where its parametric on `Long` or `Int. Some things like `LayoutDefinition` will have `RasterExtent[Long]` some things like `Raster` will have `RasterExtent[Int]` and interface between the two will be pretty infrequent in practice and require `.toInt` or `.toLong` call. Although it seems at least case classes are always covariant on primitives and will implicitly upgrade `Int` to `Long` when given (will dig into that behavior more later).

metasim:
@echeipesh I think your assessment is sound.... I classic case of trying to replace one board on the deck and ending up rebuilding it. I think the exercise brought to the fore a few ideas and considerations, but I'm concerned about it being too much of a distraction at this point. We can refer back to it if we need; how about we close it and you take the lead in how a refactoring should proceed?

echeipesh:
I'd like to keep it open for a minute for reference, I'm not dropping the work.
Although it certainly not going to get merged in this form.
Are you proposing "clean-up PRs" motion?

PR #126, opened 2019-02-25
Convert Method

jbouffard:
This PR supersedes #115 by simplifying the logic around converting `CellType`s.


vlm/src/main/scala/geotrellis/contrib/vlm/TargetCellType.scala
```diff
@@ -0,0 +1,29 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package geotrellis.contrib.vlm
+
+import geotrellis.raster.{Raster, CellType, MultibandTile}
+
+sealed trait TargetCellType {
+  // this is a by name parameter, as we don't need to call the source in all ResampleGrid types
```
echeipesh: stray c/p comment

vlm/src/main/scala/geotrellis/contrib/vlm/avro/GeotrellisRasterSource.scala
```diff
@@ -35,7 +35,12 @@ import geotrellis.raster.{MultibandTile, Tile}
   * @param layerId source layer from above catalog
   * @param bandCount number of bands for each tile in above layer
   */
-case class GeotrellisRasterSource(uri: String, layerId: LayerId, bandCount: Int = 1) extends RasterSource {
+case class GeotrellisRasterSource(
+  uri: String,
+  layerId: LayerId,
+  bandCount: Int = 1,
+  private[vlm] val targetCellType: Option[TargetCellType] = None
```
echeipesh: What are the effects of having this private?
Does scala create a whole new constructor for it or does the whole constructor becomes usable only by `vlm` package? Why do you think it should be private in general ?

vlm/src/main/scala/geotrellis/contrib/vlm/gdal/GDALBaseRasterSource.scala
```diff
@@ -93,6 +93,37 @@ trait GDALBaseRasterSource extends RasterSource {
...
+   *  would become 64.
+   *
+   *  Thus, it is recommended that one avoids converting to Byte without first
+   *  ensuring that no data will be lost.
+   *
+   *  Note:
+   *
+   *  It is not currently possible to convet to the [[BitCellType]] using GDAL.
+   *  @group convert
+   */
```
echeipesh: Nice

vlm/src/main/scala/geotrellis/contrib/vlm/avro/GeotrellisRasterSource.scala
jbouffard: From both my own understanding and [this StackOverflow answer](https://stackoverflow.com/a/14694858/8469382), it looks like making `targetCellType` `val private` is the same as making it `val` just that the getter for that value becomes `private`. I'd assume then that this getter's scope is set to the `vlm` package.

I think it should be private because it seems like a piece of plumbing that the user doesn't really need to access.

PR #130, opened 2019-03-04
Integrate GDAL Warp Bindings

jamesmcclain:
frankenstein_1818_edition_title_page

Future work:
   - [x] Test on staging
      - [x] Correctness
      - [x] Performance
   - [x] Address (intentionally) leaking tokens
   - [x] Provide access to both datasets in wrapper so that extra datasets on the Scala side can be removed
   - [x] Completely remove dependence on `geotrellis-gdal`
   - [x] Macintosh https://github.com/geotrellis/gdal-warp-bindings/pull/23
   - ~~Windows~~ https://github.com/geotrellis/gdal-warp-bindings/pull/23

See: https://github.com/geotrellis/gdal-warp-bindings

Screenshot from 2019-03-20 16_06_54


vlm/src/main/scala/geotrellis/contrib/vlm/package.scala
```diff
@@ -69,6 +75,118 @@ package object vlm {
...
+      dataType(0)
+    }
+
+    def cellSize(): CellSize = {
+      val transform = Array.ofDim[Double](6)
+      GDALWarp.get_transform(token, 0, transform)
+      CellSize(transform(1), transform(5))
+    }
+
+    def cellType(): CellType = {
```
pomadchin: In reality GeoTrellis and the TIFF lib have a bit richer types for TIFFs. 
GDAL erases some information (for instance a sign of values, [tiff tag 339](https://www.awaresystems.be/imaging/tiff/tifftags/sampleformat.html))

That's how we convert GDAL type into the GeoTrellis type (not perfect since GDAL doesn't allow us to get access to the TIFF TAG ): https://github.com/geotrellis/geotrellis-gdal/blob/b81508039d26999e8a829a2ff432e9132cb24169/gdal/src/main/scala/geotrellis/gdal/GDALUtils.scala#L39

[That's how we do it in GeoTrellis Core reader](https://github.com/locationtech/geotrellis/blob/ff27a4654f3e488c3313a8ab52069fe3f16f8f78/raster/src/main/scala/geotrellis/raster/io/geotiff/reader/GeoTiffReader.scala#L258) 

In reality in the link above `BandType` is defined by the `bits per sample tag` and `sample format`. GDAL provides a `pits per sample` thing but not a `sample format`. [Tags actually GeoTrellis reader supports](https://github.com/locationtech/geotrellis/blob/ff27a4654f3e488c3313a8ab52069fe3f16f8f78/raster/src/main/scala/geotrellis/raster/io/geotiff/tags/codes/TagCodes.scala#L77).

It can be pretty important to keep behaviour of the original reader and the new one as similar as it's possible.

vlm/src/main/scala/geotrellis/contrib/vlm/package.scala
```diff
@@ -69,6 +75,118 @@ package object vlm {
...
+        case GDALWarp.GDT_CInt16 => throw new Exception("Unsupported data type")
+        case GDALWarp.GDT_CInt32 => throw new Exception("Unsupported data type")
+        case GDALWarp.GDT_CFloat32 => throw new Exception("Unsupported data type")
+        case GDALWarp.GDT_CFloat64 => throw new Exception("Unsupported data type")
+        case _ => throw new Exception("Unknown data type")
+      }
+    }
+
+    def readTile(gb: GridBounds, band: Int): Tile = {
+      val xmin = gb.colMin
```
pomadchin: A tip: 
```scala
val GridBounds(xmin, ymin, xmax, ymax) = gb
```

vlm/src/main/scala/geotrellis/contrib/vlm/package.scala
```diff
@@ -69,6 +75,118 @@ package object vlm {
...
+        math.max(x1, x2),
+        math.max(y1, y2))
+
+      RasterExtent(e,
+        math.abs(transform(1)), math.abs(transform(5)),
+        width_height(0), width_height(1))
+    }
+
+    def resolutions: List[RasterExtent] = {
+      val N = 1 << 8;
```
pomadchin: So we're limited by 256 overviews. Mb we can move this into `Constants`? @jpolchlo had an idea to get rid of unnamed and random values from the GeoTrellis core: https://github.com/locationtech/geotrellis/issues/2872

vlm/src/main/scala/geotrellis/contrib/vlm/package.scala
```diff
@@ -69,6 +75,118 @@ package object vlm {
     new StreamingByteReader(rr)
   }
 
+  implicit class tokenMethods(val token: Long) extends AnyVal {
+
+    def getProjection(): Option[String] = {
+      val crs = Array.ofDim[Byte](1 << 16)
```
pomadchin: There is an allocation of a huge array, is it planned to hide such allocations to the C side and to return already a UTF-8 string?

vlm/src/main/scala/geotrellis/contrib/vlm/gdal/GDALDataType.scala
```diff
@@ -0,0 +1,66 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package geotrellis.contrib.vlm.gdal
+
+import com.azavea.gdal.GDALWarp
+
+
+object GDALDataType {
```
pomadchin: Mostly for @echeipesh:

I think we can merge everything into contrib now, and after that to port some things into geotrellis-gdal that should be necessarily there. What to you think? @echeipesh @jamesmcclain

vlm/src/main/scala/geotrellis/contrib/vlm/gdal/GDALUtils.scala
```diff
@@ -0,0 +1,108 @@
...
+
+package geotrellis.contrib.vlm.gdal
+
+import geotrellis.raster._
+import geotrellis.raster.io.geotiff._
+import geotrellis.raster.resample._
+
+
+object GDALUtils {
+  def deriveResampleMethodString(method: ResampleMethod): String =
```
pomadchin: The same story about splitting things into geotrellis-gdal and geotrellis-contrib. @echeipesh

vlm/src/main/scala/geotrellis/contrib/vlm/package.scala
```diff
@@ -69,6 +75,118 @@ package object vlm {
...
+      widths.zip(heights).flatMap({ case (w, h) =>
+        if (w > 0 && h > 0) Some(RasterExtent(extent, cols = w, rows = h))
+        else None
+      }).toList
+    }
+
+    def extent(): Extent = token.rasterExtent.extent
+
+    def bandCount(): Int = {
+      val count = Array.ofDim[Int](1)
```
pomadchin: The same API question is here, `get_band_count` can definitly return an Integer in this case, what do you think?

vlm/src/main/scala/geotrellis/contrib/vlm/package.scala
```diff
@@ -69,6 +75,118 @@ package object vlm {
...
+      GDALWarp.get_band_count(token, 0, count)
+      count(0)
+    }
+
+    def crs(): CRS = {
+      val crs = Array.ofDim[Byte](1 << 16)
```
pomadchin: There is again a large array allocation ):

vlm/src/test/scala/geotrellis/contrib/vlm/RasterSourceRDDSpec.scala
```diff
@@ -53,6 +54,8 @@ class RasterSourceRDDSpec extends FunSpec with TestEnvironment with BetterRaster
   val reprojectedSource = rasterSource.reprojectToGrid(targetCRS, layout)
 
   describe("reading in GeoTiffs as RDDs") {
+    GDALWarp.init(1<<8, 1<<2)
```
pomadchin: A design question: should be always known ahead of time? For instance before the application start? Also what are the rules to select this numbers?

vlm/src/test/scala/geotrellis/contrib/vlm/gdal/GDALReaderSpec.scala
```diff
@@ -1,26 +0,0 @@
...
-
-class GDALReaderSpec
-  extends FunSpec
-    with TestEnvironment
-    with RasterMatchers {
-
-  describe("GDALReaderSpec") {
-    it("should read full raster correct") {
-      val filePath = s"${new File("").getAbsolutePath()}/src/test/resources/img/aspect-tiled.tif"
-      val gdalTile = GDALReader(filePath).read().toArrayTile
```
pomadchin: I would say that it won't be bad if it would be possible to have some test that compares the GeoTiff Reader and the GDAL bindings one.

vlm/src/test/scala/geotrellis/contrib/vlm/gdal/GDALWarpOptionsSpec.scala
```diff
@@ -52,46 +55,40 @@ class GDALWarpOptionsSpec extends FunSpec with RasterMatchers with BetterRasterM
...
     val opts =
-      GDALWarpOptions(Some("MEM"))
+      GDALWarpOptions()
         .reproject(
           rasterExtent = GridExtent(Extent(630000.0, 215000.0, 645000.0, 228500.0), 10, 10).toRasterExtent,
           CRS.fromString("+proj=lcc +lat_1=36.16666666666666 +lat_2=34.33333333333334 +lat_0=33.75 +lon_0=-79 +x_0=609601.22 +y_0=0 +datum=NAD83 +units=m +no_defs "),
           WebMercator,
           ReprojectOptions.DEFAULT.copy(targetCellSize = CellSize(10, 10).some)
         )
-        .resample(
```
pomadchin: how do you test `reproject.resample` thing? Otherwise how the tests pass?

vlm/src/test/scala/geotrellis/contrib/vlm/gdal/GDALWarpOptionsSpec.scala
```diff
@@ -101,25 +98,23 @@ class GDALWarpOptionsSpec extends FunSpec with RasterMatchers with BetterRasterM
...
         GDALRasterSource(filePath)
           .reproject(
-            targetCRS        = WebMercator,
+          targetCRS        = WebMercator,
             reprojectOptions = ReprojectOptions.DEFAULT.copy(targetCellSize = CellSize(10, 10).some),
             strategy         = AutoHigherResolution
-          )
-          .resampleToRegion(
-            region = RasterExtent(Extent(-8769160.0, 4257700.0, -8750630.0, 4274460.0), CellSize(22, 22))
-          )
```
pomadchin: This is also kind of a 'stacking' operation happens here. Btw this particular example _can have a numerical mistake_ so feel free to fix it.

vlm/src/main/scala/geotrellis/contrib/vlm/package.scala
jamesmcclain: Alot of what you see here is stub code.  I want to test this before concertizing too much.  This limit is ultimately imposed by the number of available file descriptors (other factors) and not by my code.

vlm/src/main/scala/geotrellis/contrib/vlm/package.scala
jamesmcclain: Alot of what you see here is stub code.

vlm/src/main/scala/geotrellis/contrib/vlm/gdal/GDALDataType.scala
jamesmcclain: I think it needs to be tested on staging before we do any merging.

vlm/src/main/scala/geotrellis/contrib/vlm/package.scala
jamesmcclain: The return value is already occupied by a boolean.  That boolean is (intended to be used) to react to transient failures, so I am strongly in favor of keeping it around (that is true for all of these functions, they all return bool).

vlm/src/main/scala/geotrellis/contrib/vlm/package.scala
jamesmcclain: Alot of what you see here is stub code.

vlm/src/test/scala/geotrellis/contrib/vlm/RasterSourceRDDSpec.scala
jamesmcclain: Alot of what you see here is stub code.

vlm/src/test/scala/geotrellis/contrib/vlm/gdal/GDALWarpOptionsSpec.scala
jamesmcclain: This is one of the questions that I had.  I was getting a different region from merging the warp options and passing them to GDAL than this code was producing.  Removing it caused them to match -- I want to talk about that.

vlm/src/test/scala/geotrellis/contrib/vlm/gdal/GDALWarpOptionsSpec.scala
jamesmcclain: The difference that I saw did not look numerical, my recollection is that cell size was entirely different.  It could be that that was not getting properly built-in to the warp options some where.

vlm/src/main/scala/geotrellis/contrib/vlm/package.scala
jamesmcclain: Thanks

vlm/src/main/scala/geotrellis/contrib/vlm/package.scala
jamesmcclain: Okay, that is helpful, thanks.

vlm/src/main/scala/geotrellis/contrib/vlm/package.scala
jamesmcclain: Also, I don't think this limit affects the number of overviews (at least not GDAL overviews), this is to do with datasets.

vlm/src/test/scala/geotrellis/contrib/vlm/gdal/GDALWarpOptionsSpec.scala
jamesmcclain: Okay, thanks for the talk, I understand what is happening here, now.

vlm/src/test/scala/geotrellis/contrib/vlm/gdal/GDALWarpOptionsSpec.scala
jamesmcclain: Thanks, I understand what is happening now.

vlm/src/test/scala/geotrellis/contrib/vlm/gdal/GDALWarpOptionsSpec.scala
jamesmcclain: Okay, addressed.

vlm/src/test/scala/geotrellis/contrib/vlm/gdal/GDALWarpOptionsSpec.scala
jamesmcclain: Addressed.

vlm/src/main/scala/geotrellis/contrib/vlm/package.scala
jamesmcclain: Okay, using GDALUtils, now.

PR #134, opened 2019-03-08
Add MosaicRasterSource

jisantuc:
Overview
-----

This PR adds a `MosaicRasterSource` trait inheriting from `RasterSource` that represents collections of
`RasterSource`s.

Notes
-----

~failing right now because it's not merging data from the two input tifs. working on it.~ fixed in 912999d

Testing
-----

- tests

Closes raster-foundry/raster-foundry#4746


vlm/src/main/scala/geotrellis/contrib/vlm/MosaicRasterSource.scala
```diff
@@ -0,0 +1,149 @@
...
+  */
+trait MosaicRasterSource extends RasterSource {
+  // Orphan instance for semigroups for rasters, so we can combine
+  // Option[Raster[_]]s later
+  implicit val rasterSemigroup: Semigroup[Raster[MultibandTile]] =
+    new Semigroup[Raster[MultibandTile]] {
+      def combine(l: Raster[MultibandTile], r: Raster[MultibandTile]) = {
+        val targetRE = RasterExtent(
+          l.rasterExtent.extent combine r.rasterExtent.extent,
+          l.rasterExtent.cellSize)
```
moradology: I'm wondering if we shouldn't choose the smallest cell size in an attempt to preserve detail where it exists

vlm/src/main/scala/geotrellis/contrib/vlm/MosaicRasterSource.scala
```diff
@@ -0,0 +1,149 @@
...
+    * having to make sure at compile time that you're threading CRSes through everywhere correctly.
+    */
+  val commonCrs: CRS
+
+  /**
+    * Uri is required only for compatibility with RasterSource.
+    *
+    * It doesn't make sense to access "the" URI for a collection, so this throws an exception.
+    */
+  def uri: String = throw new NotImplementedError(
```
moradology: I'm wondering if this shouldn't be reflected in changes all the way up top - does that look like a heavy lift?

vlm/src/main/scala/geotrellis/contrib/vlm/MosaicRasterSource.scala
```diff
@@ -0,0 +1,149 @@
...
+    """
+      | MosaicRasterSources don't have a single uri. Perhaps you wanted the uri from
+      | one of this MosaicRasterSource's sources?
+    """.trim.stripMargin
+  )
+
+  def crs: CRS = commonCrs
```
moradology: Perhaps we can just use this in place of the extra `commonCrs`

vlm/src/main/scala/geotrellis/contrib/vlm/MosaicRasterSource.scala
```diff
@@ -0,0 +1,149 @@
...
+
+  /**
+    * The bandCount of the first [[RasterSource]] in sources
+    *
+    * If this value is larger than the bandCount of later [[RasterSource]]s in sources,
+    * reads of all bands will fail. It is a client's responsibility to construct
+    * mosaics that can be read.
+    */
+  def bandCount: Int = sources.head.bandCount
```
moradology: This is sensible enough, but we could alternatively allow different numbers of bands if no band means a full `NoData` tile

vlm/src/main/scala/geotrellis/contrib/vlm/MosaicRasterSource.scala
```diff
@@ -0,0 +1,149 @@
...
+
+  def cellType: CellType = sources.head.cellType
```
moradology: Probably want to get the LuB across all sources... really should be easier to do this from GT

vlm/src/main/scala/geotrellis/contrib/vlm/MosaicRasterSource.scala
```diff
@@ -0,0 +1,149 @@
...
+    */
+  def reproject(crs: CRS, reprojectOptions: Reproject.Options, strategy: OverviewStrategy)
+      : RasterSource = MosaicRasterSource(
+    sources map { _.reproject(crs, reprojectOptions, strategy) },
+    crs
+  )
+
+  def read(extent: Extent, bands: Seq[Int]): Option[Raster[MultibandTile]] = {
+    val rasters = sources map { _.reproject(commonCrs).read(extent, bands) }
+    rasters.reduce
```
moradology: So cool

vlm/src/main/scala/geotrellis/contrib/vlm/MosaicRasterSource.scala
```diff
@@ -0,0 +1,149 @@
...
+      sources map { _.convert(targetCellType) },
+      commonCrs
+    )
+  }
+}
+
+private object MosaicRasterSource {
+  /** Create a MosaicRasterSource with sources and bands set from the provided parameters
+    */
+  def apply(_sources: NonEmptyList[RasterSource], _commonCrs: CRS) = new MosaicRasterSource {
```
moradology: Perhaps an `unsafeFromList` variant is worth adding as well

vlm/src/main/scala/geotrellis/contrib/vlm/MosaicRasterSource.scala
jisantuc: That makes sense, are cell sizes comparable? Or have we tended to use area, or just pick a dimension and call it day?

vlm/src/main/scala/geotrellis/contrib/vlm/MosaicRasterSource.scala
jisantuc: I don't think it makes sense to remove all the way up -- for non-plural `RasterSource`s, requiring a uri is pretty reasonable. Maybe this is a case where having an extra trait for singularity will save us a big headache down the road but I'm not sure.

vlm/src/main/scala/geotrellis/contrib/vlm/MosaicRasterSource.scala
jisantuc: Allow different number of bands how? Like take the min from the collection of sources?

vlm/src/main/scala/geotrellis/contrib/vlm/MosaicRasterSource.scala
jisantuc: Is there an example of somewhere where we do this? My cell type experience so far has been a huge headache

vlm/src/main/scala/geotrellis/contrib/vlm/MosaicRasterSource.scala
moradology: Area is what typically gets used, though this is arbitrary with width vs height as I suspect you have surmised

vlm/src/main/scala/geotrellis/contrib/vlm/MosaicRasterSource.scala
moradology: I was thinking that for two rasters A (having 1 band) and B (having 3 bands), raster A's 2nd and 3rd bands would simply be `NoData` tiles sized so as to allow merger

jisantuc: Ah that's beyond `bandCount` changes though right? I don't want to stake out a strong opinion on what happens when merging incompatible(-ish) rasters here. Given the doc that it's a client's responsibility to construct reasonable collections of sources (which will sort of always be true), I'd prefer to avoid an API that suggests that the library wants to do some of the work toward figuring out the meaning of merging rasters with different numbers of bands.

vlm/src/main/scala/geotrellis/contrib/vlm/MosaicRasterSource.scala
echeipesh: This is probably going to be resolved by expanding the trait hierarchy based on following logic:

- Having URI for RasterSource is useful and reasonable when it's backed by an identifiable resource.
- Having MosaicRasterSource is useful instance of RasterSource interface
- MosaicRasterSource can not be well described by single or even a list of URIs.

=> There is shared interface and there are some traits like `DatasetRasterSource` and `MosaicRasterSource`

vlm/src/main/scala/geotrellis/contrib/vlm/MosaicRasterSource.scala
echeipesh: There is not a perfect way to handle this. [`CellType.union`](https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/CellType.scala#L73) is the closest thing, but will not resolve any differences in `NODATA` values.

vlm/src/main/scala/geotrellis/contrib/vlm/MosaicRasterSource.scala
```diff
@@ -0,0 +1,140 @@
...
+    * reads of all bands will fail. It is a client's responsibility to construct
+    * mosaics that can be read.
+    */
+  def bandCount: Int = sources.head.bandCount
+
+  def cellType: CellType = sources.head.cellType
+
+  def rasterExtent = {
+    val extents = sources map { _.reproject(crs).rasterExtent }
+    extents.tail.foldLeft(extents.head) { _ combine _ }
```
echeipesh: [`RasterExtent.combine`](https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/RasterExtent.scala#L199) I'm not sure if this is expected behavior or not in this case. However the because underlying sources may not share the pixel alignment the result of `read` calls will be fudged in a surprising way where chain of `raster.reduce(raster)` is going to resample to potentially different pixel grid at each step.

echeipesh: I think the fallout is that `MosaicRasterSource` should either be given a specific `RasterExtent` to which all input sources will be confirmed to or it should require that all sources already share pixel alignment.

PR #135, opened 2019-03-08
Add tut for RasterSource

echeipesh:


vlm/src/main/tut/RasterSource.md
```diff
@@ -0,0 +1,113 @@
...
+Current implementations exist for:
+
+- GeoTrellis GeoTiff Reader
+- GeoTrellis Indexed Layer
+- GDAL
+
+## Metadata Access
+
+Metadata will be read separately from pixels and user should expect it to be cached for repeated accessed through available fields.
+Additionally `RasterSource` should always be able to describe itself using a URI string.
```
jisantuc: `MosaicRasterSource` is in conflict with this -- should this be qualified to non-collection `RasterSource`s, now that `MosaicRasterSource`s are planned?

echeipesh: Thats a good point. I'm not sure what to do about it. `MosaicRasterSource` extending `RasterSource` is what makes it useful. Do you think it would be accurate to switch to `uri: List[String]` where the list order can be overloaded to mean z-order ?

metasim:
@echeipesh Just a thought: I've worked quite extensively with `tut` in the RasterFrames documentation, submitting PRs for a couple bug fixes and new features. It is extremely hard to work with in the internals (it was really just a quick hack by the originator), and if you think you might ever need features that go beyond what you're using now, I'd take a look at [`mdoc`](https://scalameta.org/mdoc/blog/2019/01/04/v1.2.4.html). I've not actually used it yet, not having time to port RasterFrames to it (where time would be better spent on actual documentation and not tooling), but from the documentation and a cursory look at the code, I like the design much better, and think it will ultimately replace `tut` if it can reach critical mass.

PR #138, opened 2019-03-25
Comments on RasterSource API design.

metasim:
_More of a "read request" than a "pull request"_


metasim:
Basically questions the exposure of `targetCellType` and `resampleMethod`, as well as if we could do better than `String` for encoding URIs. If you're having to do `  def resampleMethod: Option[ResampleMethod] = None` in a lot of subclasses, it says to me that the abstraction is wrong.

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSource.scala
```diff
@@ -43,7 +43,12 @@ import geotrellis.util.GetComponent
   * @groupprio reproject 2
   */
 trait RasterSource extends CellGrid with AutoCloseable with Serializable {
+  // sfitch: As with `targetCellType` below, this feels like API leakage, and seems to be captured already by
+  // the `XXXResampleRasterSource` classes.
   def resampleMethod: Option[ResampleMethod]
```
echeipesh: @metasim do you think its ever valuable to know if you're working with original or resample/reprojected data? If so relying on `XXXResampledRasterSource` would require a `match` on all possible instances of it. If not, then yeah, this is leaky.

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSource.scala
```diff
@@ -43,7 +43,12 @@ import geotrellis.util.GetComponent
...
+
+  // sfitch: what if we have `RasterSource`s that aren't identified by a URI? e.g. `InMemoryRasterSource`?
+  // Also: should we consider more strongly typing this?
```
echeipesh: :+1: I think we're in for something like:

```scala 
 trait NamedRasterSource { self: RasterSource => 
   def uri: ???
}
```

`MosaicRasterSource` is another thing that doesn't have a URI sadly.

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSource.scala
metasim: @echeipesh From a purists perspective, if you do  then the abstraction is wrong. It's an implementation detail that shouldn't affect how you treat the abstraction called "RasterSource". But in essence, it's The Expression Problem, and abstracting in one direction requires specificity in another. On balance, I'd rather move that specificity away from the public API into the internal implementation.

metasim: This actually leads me to another point which is RasterSource currently mixes (at least) two concerns that may deserve to be separated:

1. How to read (sub-)rasters (efficiently)
2. How to do that while efficiently transforming the data into the desired form without large data copies

(Do you agree? Have I missed any?)

The first is obvious and core to the abstraction. The second is technically a performance optimization, but a required one. My gut says this responsibility could be separately abstracted, using something akin to an AST (maybe even a linear one), providing easier testing, reusability, and less implementation bleeding.

metasim: Something like this:

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSource.scala
metasim: @pomadchin @echeipesh Can one of you remind me of which cases won't allow us to use `java.net.URI`?

docs/RasterSource.puml
```diff
@@ -0,0 +1,67 @@
+@startuml
```
metasim: If you're not aware of it, IntelliJ has a nice PlantUML plugin that will render the results while you edit.

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSource.scala
echeipesh: S3 URIs allow characters that are not allowed by `java.net.URI`. We ran into this problem when using pre-signed URIs in Raster Foundry as well as having `|` be part of the S3 prefix for some reason :(

metasim: Is this going too far?:

```scala
trait IdentifiedRasterSource { self: RasterSource =>
  def id: RasterSourceIdentifier
}

trait RasterSourceIdentifier {
  def connect: ???
  def asString: String
}

case class PathIdientifier(path: String) extends RasterSourceIdentifier
case class URIIdientifier(uri: URI) extends RasterSourceIdentifier
import com.amazonaws.services.s3.AmazonS3URI
case class S3URIIdentified(s3URI: AmazonS3URI) extends RasterSourceIdentifier

```

If so, I'd suggest this instead:

```scala
trait IdentifiedRasterSource { self: RasterSource =>
  def id: RasterSourceIdentifier
}

// Tagged type pattern...
trait ValidRSID
type RasterSourceIdentifier = String with ValidRSID
object RasterSourceIdentifier {
  def apply(path: String): RasterSourceIdentifier = {
    if (/* path is valid URI or S3 URI */) 
      path.asInstanceOf[RasterSourceIdentifier]
    else
      // Do something obnoxious....
  }
}
```

echeipesh:
Broke out the discussion into two issues:
https://github.com/geotrellis/geotrellis-contrib/issues/198 (adding this to implementation queue)
https://github.com/geotrellis/geotrellis-contrib/issues/197 (needs more discussion I feel)

PR #140, opened 2019-03-26
GridExtent Refactor

echeipesh:
Refactor to allow `RasterSource` to represent layers that cover a tiled geotrellis layer at high resolution.

Connects: https://github.com/locationtech/geotrellis/pull/2886

### DIFF
--- a/testkit/src/main/scala/geotrellis/contrib/testkit/BetterRasterMatchers.scala
+++ b/testkit/src/main/scala/geotrellis/contrib/testkit/BetterRasterMatchers.scala
@@ -13,23 +13,24 @@ import geotrellis.raster.render.ascii._
 
 import matchers._
 import org.scalatest.tools.BetterPrinters
-import spire.syntax.cfor._
 
 import scala.reflect._
 import java.nio.file.{Files, Paths}
 import spire.math.Integral
+import spire.implicits._
 
 trait BetterRasterMatchers { self: Matchers with FunSpec with RasterMatchers =>
   import BetterRasterMatchers._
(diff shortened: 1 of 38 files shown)


echeipesh:
Finding that it is a little harder to replace `RasterExtent` with `GridExtent[Int]`. The type alias does not allow the use of all the constructors, requiring wide-sweeping code changes in `geotrellis-*` I'd like to avoid for now.

testkit/src/main/scala/geotrellis/contrib/testkit/BetterRasterMatchers.scala
```diff
@@ -48,7 +50,7 @@ trait BetterRasterMatchers { self: Matchers with FunSpec with RasterMatchers =>
   def assertTilesEqual(actual: MultibandTile, expected: MultibandTile): Unit = {
     actual should have (
       cellType (expected.cellType),
-      dimensions (expected.dimensions),
+      // dimensions (expected.dimensions),
```
pomadchin: Hm, why is it commented out?

echeipesh: Type checking on that was not working out well. The exact reason is to do with how ScalaTest matchers are being constructed, where the thing being checked is in separate param block from the thing doing the checker. Now that dimensions of `CellGrid[N]` are `N` I had hell of a time getting that matcher to work. Basically something I bailed on to come back and re-evaluate how/if those kind of matchers are worth while.

Issue #55, opened 2018-12-03
Trace level logging for reads


jisantuc:
With read errors, it's difficult to tell what's actually going on in any client application using geotrellis-contrib.

With `trace` level logging of:

- uri
- warp options
- extent

that would be easier

metasim:
+1

metasim:
Ideally would be great to know all the scenarios where GDAL is making IO calls.

metasim:
https://github.com/geotrellis/gdal-warp-bindings/issues/34

echeipesh:
Added this gem when debugging new OGC services:

https://github.com/geotrellis/geotrellis-contrib/blob/e0fc7d6b4d3cbe4d10c16c304a91a31b12a18d1d/vlm/src/main/scala/geotrellis/contrib/vlm/avro/GeotrellisReprojectRasterSource.scala#L66-L77

it has been invaluable and ANSI colors are Purdy

PR #147, opened 2019-04-01
Added ability to register a read callback on GeoTiff RangeReaders

metasim:
Helps with debugging header-only metadata and range reads. Also opens the door for the use of Spark accumulators.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -21,6 +21,7 @@ lazy val commonSettings = Seq(
     "-language:postfixOps",
     "-language:existentials",
     "-language:experimental.macros",
+//    "-Ywarn-unused-import",
     "-Ypartial-unification" // Required by Cats
   ),
   publishMavenStyle := true,
--- a/gdal/src/test/scala/geotrellis/contrib/vlm/gdal/GDALRasterSourceSpec.scala
+++ b/gdal/src/test/scala/geotrellis/contrib/vlm/gdal/GDALRasterSourceSpec.scala
@@ -17,19 +17,16 @@
 package geotrellis.contrib.vlm.gdal
 
 import geotrellis.contrib.vlm._
-import geotrellis.proj4._
 import geotrellis.raster._
 import geotrellis.raster.io.geotiff.reader.GeoTiffReader
 import geotrellis.raster.io.geotiff.AutoHigherResolution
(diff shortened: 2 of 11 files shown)


metasim:
Kinda-sorta helps with this, but not as low level. 

https://github.com/geotrellis/geotrellis-contrib/issues/55

metasim:
Is there a way to refactor to cover all `RasterSource` types?

echeipesh:
@metasim Do you think this is the type of thing that should handled through RangeReader SPI or do you think this is a type of feature we should expose in the constructor? I've seen you use this debug trick before and it seems mega helpful but I'm not clear if this is debugging-level feature or user-level feature.

I'm not sure how to refactor other RasterSource types. GeoTrellisRasterSource for instance is tiled, so nothing there is expressed as byte ranges. The new GDALRasterSources don't even expose this information and probably can't because there is non-trivial GDAL machinery backing them.

metasim:
@echeipesh All good questions, none of which I have a good answer. It's been super helpful to-date in RasterFrames pre-RasterSource, and this was one attempt at adding it back in, but is of questionable usefulness if we can't provide a general yet unobtrusive way of doing it across all RasterSources.

echeipesh:
We recently added `RangeReader` SPI:
https://github.com/locationtech/geotrellis/blob/master/util/src/main/scala/geotrellis/util/RangeReaderProvider.scala

https://github.com/locationtech/geotrellis/commit/52689bb3b558d620558162dda80320cf77ef4533#diff-2cc4bdaf59af25e9a1f0fe72ae346193

I think that gives a way to register some that does callback ... and log it in some kind of global object I suppose.

Also this Issue https://github.com/geotrellis/geotrellis-contrib/issues/155 will result in ability to provide the `GeoTiff[T]` directly to the constructor, which would allow injecting whichever kind of `RangeReader` programmatically. 

Using accumulators seems appealing, I know that we should have used them in `LayerWriter` in order to generate update histogram while we write layers but I'm not clear on how that could be a general thing for `RasterSource` interface.

echeipesh:
I'd like to close this PR because its not easily merg-able and I'm not sure what to say concretely to make it so. As always, please re-open or re-issue if I'm wrong in doing so.

PR #157, opened 2019-04-17
Implement sparse stitch to properly infer missing tiles

moradology:
This PR implements a sparse stitch function to avoid losing information in case `SpatialKeys` between tiles of imagery are missing. Without this fix, we're susceptible to two bugs if a geotrellis layer has been ingested in pieces (and thus lacks tiles between ingest locations):
1. If an extent is requested which ranges over a meaningful piece of data, that data is liable to be shifted to the top-left in case there are missing tiles prior in row-major order
2. If an extent is requested within the overall extent of a layer but in which there are no tiles, a reader will report that no tile could be grabbed (rather than an empty tile, as one might expect)

Note: I've also bumped the version to be in line with sbt-release convention

### DIFF
--- a/gdal/version.sbt
+++ b/gdal/version.sbt
@@ -1,1 +1,1 @@
-version in ThisBuild := "3.11.0"
+version in ThisBuild := "3.12.0-SNAPSHOT"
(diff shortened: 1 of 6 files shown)


pomadchin (approved):
G2G once travis is happy

pomadchin:
It is related to the GeoTrellis issue: https://github.com/locationtech/geotrellis/issues/2903

PR #159, opened 2019-04-18
Move slick project from core

moradology:
Slick is being removed from the core library as we don't want to encourage anyone to use it and its functionality is not central to GeoTrellis. We'll keep it here, in `geotrellis-contrib`, instead

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,8 @@ project/plugins/target
 project/target
 target
 .ensime
+*.bloop
+*.metals
 \#*#
 *~
 .#*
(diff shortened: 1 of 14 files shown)


project/.bloop/geotrellis-contrib-build.json
```diff
@@ -0,0 +1,1844 @@
+{
```
pomadchin: should not be there, can you add it into .gitignore?

pomadchin (changes requested):

pomadchin (approved):

PR #160, opened 2019-04-18
Test case where GDALRasterSource extent doesn't match one read by GeoTiffRasterSource

metasim:


metasim:
Trying to further establish cause.

metasim:
So far, but only happens in test file. Looks like it has something to do with the difference between `geotrellis.contrib.vlm.gdal.GDALDataset#extent` using `GDALWarp.WARPED` vs `GDALWarp.SOURCE`:

```scala
val path = VSIPath(Resource.path("img/extent-bug.tif")).vsiPath
val expected = Extent(703986.502389, 4249551.61978, 709549.093643, 4254601.8671)
val ds = GDALDataset(path)
assertSimilarExtents(ds.extent(GDALWarp.SOURCE), expected)
assertSimilarExtents(ds.extent(GDALWarp.WARPED), expected)
```

(second one fails).

pomadchin:
Hey @metasim, this PR really breaks the entire files structures and moves all files that were used for unit tests into some other folder. There are two ways to resolve it: 

* to rebase on top of master and fix tests
* this test is enough https://github.com/geotrellis/geotrellis-contrib/pull/165

metasim:
I was trying to DRY with the test files. Would you rather I leave them alone and just have copies of the test files?

metasim:
PS: If you don't think this is relevant any more, feel free to just close.

pomadchin:
Addressed by https://github.com/geotrellis/geotrellis-contrib/pull/165

@metasim feel free to create a new PR with more tests. 

The new logic is:
* to use a GeoTiff driver by default (without any resampling / reprojection / cell type chamges)
* any warp transformation would require a cellSize (we're calculating it anyway, so the only thing that was changed to pass the old cellSize into the warp options for a proper cellType conversion)

PR #167, opened 2019-05-03
Round instead of ceil for combining raster extents

jisantuc:
This PR addresses a bug in MosaicRasterSource where sometimes extents won't end up combinable after `ceil`.

Working on testing instructions, just leaving this here for a minute


echeipesh:
This makes sense on its own, its consistent with checks in `GridExtent`

https://github.com/locationtech/geotrellis/blob/fb6f94077f932f9b60ff0a5a8c97ec66b77040dc/raster/src/main/scala/geotrellis/raster/GridExtent.scala#L48-L51

PR #186, opened 2019-06-06
Hadoop ByteReader

### LINKED ISSUE #142: Implement support for HDFS
Perhaps covered by #10? But in the short term, would it be possible to support HDFS here?:

https://github.com/geotrellis/geotrellis-contrib/blob/1701c1ae3f3d37b4732c722a5a30096ec51b9de2/vlm/src/main/scala/geotrellis/contrib/vlm/package.scala#L35-L69

moradology:
# Overview

This PR adds support for loading a Hadoop `ByteReader` when the provided URI has the authority of `hdfs://`.

## Notes

Instead of attempting to lay out all possible Hadoop configurations (a doomed task), the `geotrellis-contrib` `vlm` configuration optionally includes a list of paths to XML files which is used to generate a Hadoop `Configuration`

## Checklist

Closes #142

### DIFF
--- a/vlm/src/main/resources/reference.conf
+++ b/vlm/src/main/resources/reference.conf
@@ -3,8 +3,4 @@ vlm.geotiff ={
     allow-global-read: false
     region: "us-east-1"
   }
-
-  hdfs = {
-    resources = ["file:///etc/hadoop/conf/core-site.xml", "file:///etc/hadoop/conf/hdfs-site.xml"]
-  }
 }
--- a/vlm/src/main/resources/reference.conf
+++ b/vlm/src/main/resources/reference.conf
@@ -4,5 +4,7 @@ vlm.geotiff ={
     region: "us-east-1"
   }
 
-  hdfs = ["file:///etc/hadoop/conf/core-site.xml", "file:///etc/hadoop/conf/hdfs-site.xml"]
+  hdfs = {
+    resources = ["file:///etc/hadoop/conf/core-site.xml", "file:///etc/hadoop/conf/hdfs-site.xml"]
+  }
 }
--- a/vlm/src/main/resources/reference.conf
+++ b/vlm/src/main/resources/reference.conf
@@ -1,4 +1,8 @@
-vlm.geotiff.s3 {
-  allow-global-read: false
-  region: "us-east-1"
+vlm.geotiff ={
+  s3 = {
+    allow-global-read: false
+    region: "us-east-1"
+  }
+
+  hdfs = ["file:///etc/hadoop/conf/core-site.xml", "file:///etc/hadoop/conf/hdfs-site.xml"]
 }
--- a/vlm/src/main/scala/geotrellis/contrib/vlm/config/HdfsConfig.scala
+++ b/vlm/src/main/scala/geotrellis/contrib/vlm/config/HdfsConfig.scala
@@ -21,10 +21,12 @@ import org.apache.hadoop.fs.Path
 import pureconfig.generic.auto._
 import pureconfig._
 
-case class HdfsConfig(resources: List[Path]) {
+import java.net.URI
+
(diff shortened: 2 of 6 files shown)


vlm/src/main/scala/geotrellis/contrib/vlm/config/HdfsConfig.scala
```diff
@@ -0,0 +1,39 @@
...
+    conf
+  }
+}
+
+object HdfsConfig {
+  implicit val hadoopPathReader = ConfigReader.fromString[Path] {
+    ConvertHelpers.catchReadError(pathString => new Path(pathString))
+  }
+
+  lazy val conf: HdfsConfig = pureconfig.loadConfigOrThrow[HdfsConfig]("vlm.geotiff.hdfs")
```
jbouffard: If I did `HdfsConfig.load()` would I get the same thing as `HdfsConfig.conf`?

jbouffard (changes requested):
It'd be good if there was a unit test that confirms `hdfs` `URI`s can be read.

vlm/src/main/scala/geotrellis/contrib/vlm/config/HdfsConfig.scala
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+package geotrellis.contrib.vlm.config
+
+import org.apache.hadoop.conf.Configuration
+import org.apache.hadoop.fs.Path
+import pureconfig.generic.auto._
+import pureconfig._
+
+case class HdfsConfig(resources: List[Path]) {
```
pomadchin: Just a notice that we should consider: can you use `URI` or `Strings` here? `org.apache.hadoop.fs.Path` is not serializable. It is resolved only in Hadoop 3.x: https://issues.apache.org/jira/browse/HADOOP-13519

vlm/src/main/scala/geotrellis/contrib/vlm/config/HdfsConfig.scala
moradology: `HdfsConfig.conf` will provide an instance of `HdfsConfig`; calling `load` on that object will yield a Hadoop configuration

jbouffard:
Would there be anyway to bring `HdfsConfig` over GeoTrellis once the spark split is merged? I need it for the `RangeReader` SPI.

jbouffard (approved):

pomadchin:
@jbouffard I think SPI would address this issue in a more generic way and we won't be using this `getByteReader` function anymore.

vlm/src/main/resources/reference.conf
```diff
@@ -1,4 +1,10 @@
...
-  allow-global-read: false
-  region: "us-east-1"
+vlm.geotiff ={
+  s3 = {
+    allow-global-read: false
+    region: "us-east-1"
+  }
+
+  hdfs = {
+    resources = ["file:///etc/hadoop/conf/core-site.xml", "file:///etc/hadoop/conf/hdfs-site.xml"]
```
pomadchin: mb it should be empty by default? I don't expect everyone willing to set it up explicitly. We can move it into the `test/resource/application.conf` as a usage example.

pomadchin (changes requested):
Just a single question about the default `reference.conf`

vlm/src/main/resources/reference.conf
moradology: Yeah, this seems reasonable to me.

vlm/src/main/scala/geotrellis/contrib/vlm/config/HdfsConfig.scala
moradology: I've made `conf` private to avoid exposing a potentially confusing property

PR #188, opened 2019-06-17
URI, URL, VSI Path Abstraction

### LINKED ISSUE #156: Design abstraction over URI, VSIPath, AmazonS3URI, etc.
Consider constructing an abstraction that validates and differentiates the various forms an "enhanced" URI might take. Continuing to be "stringly-typed" seems unsafe.

See also: 

* [Motivating Example](https://github.com/geotrellis/geotrellis-contrib/issues/10)
* [Related Discussion](https://github.com/geotrellis/geotrellis-contrib/pull/138#discussion_r269644737)

jbouffard:
# Overview

This PR adds the `DataPath` type which provides an abstraction for a path to be read.

## Checklist

Closes #156


gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALDataPath.scala
```diff
@@ -0,0 +1,30 @@
+package geotrellis.contrib.vlm.gdal
```
echeipesh: License headers

vlm/src/main/scala/geotrellis/contrib/vlm/DataPath.scala
```diff
@@ -0,0 +1,15 @@
...
+
+trait DataPath {
+  def path: String
+  def targetPath: String
+
+  protected def servicePrefix: String
+
+  override def toString: String = targetPath
+  def toURI: URI = new URI(targetPath)
+  def toURL: URL = new URL(targetPath)
```
echeipesh: We should avoid using `URL` in API because it's not serializable and binds to resources in surprising ways. This is "well known" java API issue.

vlm/src/main/scala/geotrellis/contrib/vlm/DataPath.scala
```diff
@@ -0,0 +1,15 @@
+package geotrellis.contrib.vlm
+
+import java.net.{URI, URL}
+
+
+trait DataPath {
+  def path: String
+  def targetPath: String
+
+  protected def servicePrefix: String
```
echeipesh: This is not part of public interface, so its not useful to use and is not used on the base trait, so for now this unnecessarily restricts the possible subclasses of `DataPath`. Either it should become somehow public and documented or be removed.

vlm/src/main/scala/geotrellis/contrib/vlm/DataPath.scala
```diff
@@ -0,0 +1,15 @@
+package geotrellis.contrib.vlm
+
+import java.net.{URI, URL}
+
+
+trait DataPath {
```
echeipesh: Doc-strings for public interfaces are helpful.

gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALDataPath.scala
```diff
@@ -0,0 +1,30 @@
...
+
+import geotrellis.contrib.vlm.DataPath
+
+import java.net.MalformedURLException
+
+
+case class GDALDataPath(val path: String) extends DataPath {
+  protected def servicePrefix: String = "gdal+"
+
+  private lazy val strippedPath: String =
```
echeipesh: `lazy` is to be avoided without specific reason. here it might delay error checking on the URI until its used somehow which is unhelpful behavior.

gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALDataPath.scala
```diff
@@ -0,0 +1,30 @@
...
+    if (path.startsWith(servicePrefix))
+      path.splitAt(servicePrefix.size)._2
+    else
+      path
+
+  private lazy val vsiPath: VSIPath = VSIPath(strippedPath)
```
echeipesh: Can `GDALDataPath` and `VSIPath` be the same thing? They're both basically "case class that is understood by GDAL layer and handles parsing and checking".

gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALRasterSource.scala
```diff
@@ -137,3 +121,11 @@ case class GDALRasterSource(
     readBounds(bounds, 0 until bandCount)
   }
 }
+
+object GDALRasterSource {
+  def apply(path: String): GDALRasterSource =
+    GDALRasterSource(GDALDataPath(path))
+
+  def apply(path: String, options: GDALWarpOptions): GDALRasterSource =
+    GDALRasterSource(GDALDataPath(path), options)
```
echeipesh: We've gone heavy on `apply` overloads in the past and it produced a lot of indirection. I wonder if having implicit `String => GDALDataPath` in the companion object of `GDALDataPath` would be better? It would at least cover the interactive case and would replace a lot of `apply` methods.

@moradology @pomadchin

gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALDataPath.scala
pomadchin: For the source code use `createHeaders` and `test:createHeaders` to generate headers in tests.

vlm/src/main/scala/geotrellis/contrib/vlm/avro/GeoTrellisDataPath.scala
```diff
@@ -0,0 +1,14 @@
+package geotrellis.contrib.vlm.avro
+
+import geotrellis.contrib.vlm.DataPath
+
+
+case class GeoTrellisDataPath(val path: String) extends DataPath {
```
echeipesh:
```scala
val rs: RasterSource = RasterSource("gt+s3://azavea-datahub/catalog") 
// how do I specify the Layer name or zoom? I don't have enough information here.
```


## Option 1
Try to parse out the parameters from path
`gt+s3://azavea-datahub/catalog/nlcd` -> pick the best zoom level
`gt+s3://azavea-datahub/catalog/nlcd/12` -> pick the zoom level I told you

## Option 2
Use the query parameters
`gt+s3://azavea-datahub/catalog?layer=nlcd`
`gt+s3://azavea-datahub/catalog?layer=nlcd&zoom=12`

gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALDataPath.scala
```diff
@@ -0,0 +1,30 @@
...
+
+  def targetPath: String =
```
echeipesh: I don't think this is part of public interface either. The parameter `path` can be used in `toString` and each individual subclass can have a field that represented the parsed/transformed value like `.vsiPath` for GDAL and `.catalogUri` and `.layerName` for GeoTrellis catalog etc.

pomadchin:
@jbouffard Also a side question, have you considered https://github.com/lemonlabsuk/scala-uri? Though it's usage is a bit blocked by https://github.com/lemonlabsuk/scala-uri/pull/38

// @echeipesh

gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALRasterSource.scala
jbouffard: After thinking about it, I feel that these types of `apply`s are okay. They don't perform any real logic, and they return a `GDALRasterSource` in their body instead of calling another `apply`. I think `apply` methods like these are fine since they're essentially the same as an implicit method - all they do is provide a some convenience. It's when there's actual work being done in the `apply`s and they start calling each other that problems start emerging.

With the above being said, I wouldn't be opposed to using implicits, but I also don't think there's anything wrong with the current code.

pomadchin: @echeipesh I would like to try a new approach, we can always revert back to overloads. The approach with `String => Path` implicit with have in `HadoopGeoTiffRDD.read{Singleband|Multiband}` overloads, the case is that we already have too many overloads and we wanted everything to work not only with `Path`  but also with `String`: https://github.com/locationtech/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/store/hadoop/HadoopGeoTiffRDD.scala#L116 and implicit method is for some reason here: https://github.com/locationtech/geotrellis/blob/master/store/src/main/scala/geotrellis/store/hadoop/package.scala#L24

Even though it is not in a companion object - it never created problems (since it is a pretty straightforward implicit) and it made the code much more transparent and easier to read (as it reduced overloads number two times).

jbouffard:
@pomadchin Thanks for suggesting that library. I'll take a look into it :+1:

pomadchin:
@jbouffard 1.4.7 version is released, it is without `spray-json` dependency!

vlm/src/main/scala/geotrellis/contrib/vlm/avro/GeoTrellisDataPath.scala
```diff
@@ -0,0 +1,71 @@
...
+ *      - '''band_count''': The number of bands of each Tile in the layer.
+ *    Of the above three parameters, `layer` and `zoom` are required. In addition,
+ *    this path can be prefixed with, '''gt+''' to signify that the target path
+ *    is to be read in only by [[GeotrellisRasterSource]].
+ *  @example "s3://bucket/catalog?layer=layer_name&zoom=10"
+ *  @example "hdfs://data-folder/catalog?layer=name&zoom-12&band_count=5"
+ *  @example "gt+file:///tmp/catalog?layer=name&zoom=5"
+ *  @note The order of the query parameters does not matter.
+ */
+case class GeoTrellisDataPath(val path: String) extends DataPath {
```
jbouffard: I want to point out here that the files/types that I've added concerning GT are spelled differently than other files in the same package. Mine are spelled as: `GeoTrellis` whereas the other files are: `Geotrellis`. I don't feel too strongly about either or, but `GeoTrellis` is the correct spelling, so I wasn't sure if that's something we'd want to change to.

vlm/src/main/scala/geotrellis/contrib/vlm/avro/GeoTrellisDataPath.scala
```diff
@@ -0,0 +1,71 @@
...
+  private val servicePrefix: String = "gt+"
+
+  private val layerNameParam: String = "layer"
+  private val zoomLevelParam: String = "zoom"
+  private val bandCountParam: String = "band_count"
+
+  require(path.contains(s"$layerNameParam="), s"The layer query parameter must be in the given path: $path")
+
+  // TODO: Support having the zoom parameter be optional
+  require(path.contains(s"$zoomLevelParam="), s"The zoom query parameter must be in the given path: $path")
```
jbouffard: I wasn't sure if finding an appropriate zoom level if none was given fit in this PR or not, so I made it a requirement for right now.

vlm/src/main/scala/geotrellis/contrib/vlm/avro/GeoTrellisDataPath.scala
```diff
@@ -0,0 +1,71 @@
...
+
+  private val strippedPath: String =
+    if (path.startsWith(servicePrefix))
+      path.splitAt(servicePrefix.size)._2
+    else
+      path
+
+  val Array(catalogPath, queryParams) = strippedPath.split('?')
+
+  private val brokenUpParams: Array[(String, String)] =
```
jbouffard: @pomadchin mentioned that there's a library called, [scala-uri](https://github.com/lemonlabsuk/scala-uri) that would handle the parsing for us; and it seems to do exactly what we're looking for. However, I'm not sure if it's worth bringing in another dependency to do this right now. If we start adding more complexity to the `URI`s, then I think that library should be something to consider.

pomadchin: I would say that it would be pretty good to get rid of all that awful amount of regexps that make these `DataPath` classes unreadable. I created an issue for it: https://github.com/geotrellis/geotrellis-contrib/issues/192 

We can parse 1. schemes better 2. query params 3. authority 4. paths 5. encodes everything and supports lots of charsets. Basically we can get rid of the most `findFirstIn` conversions and lots of `requires` in code that break all the functional style of our library.

vlm/src/main/scala/geotrellis/contrib/vlm/avro/GeoTrellisDataPath.scala
pomadchin: I'm worrying a lot about such an enormous amount of `require` calls everywhere, we have the same issue in GeoTrellis. I would like not at least to increase amount of these `require` calls.

vlm/src/main/scala/geotrellis/contrib/vlm/avro/GeoTrellisDataPath.scala
pomadchin: `GeoTrellis` is a good thing, let's standardize on it

gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALDataPath.scala
```diff
@@ -16,18 +16,80 @@
...
+ *    `VSI Format`. Thus, if given another format type, this class will format it
+ *    so that it can be read.
+ *
+ *  @param compressedFileDelimiter `String` used to represent a file that's to be read
+ *    from a compressed.
+ *  @example "zip+s3://bucket/prefix/zipped-data.zip!data.tif"
  */
-case class VSIPath(uri: String, compressedFileDelimiter: String = "!") {
+case class GDALDataPath(
+  val path: String,
```
pomadchin: It is already public, `val` should be removed from here.

gdal/src/test/scala/geotrellis/contrib/vlm/gdal/GDALDataPathSpec.scala
```diff
@@ -29,55 +29,58 @@ class VSIPathSpec extends FunSpec with Matchers {
...
 
       it("zip+http url") {
         val filePath = "www.radomdata.com/test-files/data.zip"
         val url = s"zip+http://$filePath"
-        val expectedPath = s"/vsizip//vsicurl/$url"
+        val expectedPath = s"/vsizip//vsicurl/http://$filePath"
 
-        VSIPath(url).vsiPath should be (expectedPath)
+        println(s"\n\nexpectedPath: $expectedPath")
+        println(s"actualPath: ${GDALDataPath(url).vsiPath}\n\n")
```
pomadchin: Should it be removed?

jbouffard: Whoops, yeah they should be.

gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALRasterSourceProvider.scala
```diff
@@ -0,0 +1,33 @@
...
+package geotrellis.contrib.vlm.gdal
+
+import geotrellis.contrib.vlm.RasterSourceProvider
+
+
+class GDALRasterSourceProvider extends RasterSourceProvider {
+  def canProcess(path: String): Boolean =
+    try {
+      GDALDataPath(path)
+      return true
```
pomadchin: `return` is not required here

vlm/src/main/scala/geotrellis/contrib/vlm/avro/GeoTrellisDataPath.scala
```diff
@@ -0,0 +1,87 @@
...
+
+  /** The zoom level of the target layer */
+  val zoomLevel: Option[Int] = mappedQueryParams.get(zoomLevelParam)
+
+  /** The band count of the target layer */
+  val bandCount: Option[Int] = mappedQueryParams.get(bandCountParam)
+}
+
+object GeoTrellisDataPath {
+  implicit def toGeoTrellisDataPath(path: String): GeoTrellisDataPath =
```
pomadchin: :+1:

vlm/src/main/scala/geotrellis/contrib/vlm/avro/GeoTrellisRasterSourceProvider.scala
```diff
@@ -0,0 +1,45 @@
...
+
+import geotrellis.contrib.vlm._
+import geotrellis.spark.LayerId
+
+
+class GeoTrellisRasterSourceProvider extends RasterSourceProvider {
+  def canProcess(path: String): Boolean =
+    try {
+      GeoTrellisDataPath(path)
+      return true
```
pomadchin: [return](https://tpolecat.github.io/2014/05/09/return.html) is used in some very specific and performance sensitive java-like operations.

vlm/src/main/scala/geotrellis/contrib/vlm/avro/GeoTrellisRasterSourceProvider.scala
```diff
@@ -0,0 +1,45 @@
...
+    } catch {
+      case _: Throwable => false
+    }
+
+    def rasterSource(path: String): GeotrellisRasterSource = {
+      val dataPath = GeoTrellisDataPath(path)
+      val layerName = dataPath.layerName
+
+      // TODO: Handle the case where there's no zoom level
```
pomadchin: Do you need to fix this TODO before merging?

vlm/src/main/scala/geotrellis/contrib/vlm/avro/GeotrellisRasterSource.scala
```diff
@@ -26,6 +26,9 @@ import geotrellis.spark.io._
 import geotrellis.spark.{LayerId, Metadata, SpatialKey, TileLayerMetadata}
 import geotrellis.vector._
 
+import java.net.URI
```
pomadchin: Why do you need this import here?

vlm/src/main/scala/geotrellis/contrib/vlm/geotiff/GeoTiffRasterSourceProvider.scala
```diff
@@ -0,0 +1,33 @@
...
+package geotrellis.contrib.vlm.geotiff
+
+import geotrellis.contrib.vlm._
+
+
+class GeoTiffRasterSourceProvider extends RasterSourceProvider {
+  def canProcess(path: String): Boolean =
+    try {
+      GeoTiffDataPath(path)
+      return true
```
pomadchin: return here again

vlm/src/test/scala/geotrellis/contrib/vlm/avro/GeotrellisRasterSourceSpec.scala
```diff
@@ -31,9 +31,11 @@ import geotrellis.vector.Extent
 import org.scalatest.{FunSpec, GivenWhenThen}
 
 class GeotrellisRasterSourceSpec extends FunSpec with RasterMatchers with BetterRasterMatchers with GivenWhenThen with CatalogTestEnvironment {
-  val uriMultiband = s"file://${TestCatalog.multibandOutputPath}"
-  val uriSingleband = s"file://${TestCatalog.singlebandOutputPath}"
   val layerId = LayerId("landsat", 0)
+  val params = s"?layer=${layerId.name}&zoom=${layerId.zoom}"
+  val uriMultibandNoParams = s"file://${TestCatalog.multibandOutputPath}"
+  val uriMultiband = s"file://${TestCatalog.multibandOutputPath}$params"
```
pomadchin: Why do we need such URIs here, since the dispatch by layerId happens inside the `GeoTrellisRasterSource` (`LayerId` is passed as an extra constructor parameter).

vlm/src/test/scala/geotrellis/contrib/vlm/spark/RasterSummarySpec.scala
```diff
@@ -92,7 +92,7 @@ class RasterSummarySpec extends FunSpec with TestEnvironment with BetterRasterMa
   }
 
   it("should create ContextRDD from RDD of GeoTiffRasterSources") {
-    val inputPath = Resource.path("img/aspect-tiled.tif")
+    val inputPath = GeoTiffDataPath(Resource.path("img/aspect-tiled.tif"))
```
pomadchin: It is handled via implicit `String => GeoTiffDataPath` conversion?

vlm/src/main/scala/geotrellis/contrib/vlm/avro/GeotrellisRasterSource.scala
```diff
@@ -109,14 +112,14 @@ class GeotrellisRasterSource(
...
+    new GeotrellisResampleRasterSource(attributeStore, dataPath, closestLayerId, sourceLayers, resampledGridExtent, method, targetCellType)
   }
 
   def convert(targetCellType: TargetCellType): RasterSource =
-    new GeotrellisRasterSource(attributeStore, uri, layerId, sourceLayers, bandCount, Some(targetCellType))
+    new GeotrellisRasterSource(attributeStore, dataPath, layerId, sourceLayers, bandCount, Some(targetCellType))
 
   override def toString: String =
-    s"GeoTrellisRasterSource($uri,$layerId)"
+    s"GeoTrellisRasterSource(${dataPath.toString},$layerId)"
```
pomadchin: TIP: you already overrided to string, in the string interpolator `toString` method would be invoked to convert object into string already.

vlm/src/main/scala/geotrellis/contrib/vlm/avro/GeoTrellisRasterSourceProvider.scala
jbouffard: I'm not sure if I do, since I know that's a feature we want that's not currently implemented (it seemed to be beyond the scope of this PR).

vlm/src/test/scala/geotrellis/contrib/vlm/avro/GeotrellisRasterSourceSpec.scala
jbouffard: The `uriMultibandNoParams` val is for the `CollectionLayerReader` since that can't take `URI`s with query parameters.

pomadchin: @jbouffard the question is in this:
```scala
val uriMultiband = s"file://${TestCatalog.multibandOutputPath}$params"
val uriSingleband = s"file://${TestCatalog.singlebandOutputPath}?layer=${layerId.name}&zoom=${layerId.zoom}"

lazy val sourceMultiband = new GeotrellisRasterSource(uriMultiband, layerId)
lazy val sourceSingleband = new GeotrellisRasterSource(uriSingleband, layerId)
```

Do we need such formated uris for `GeoTrellisRasterSources`? It looks like it doesn't use it at all. I though such URI is required for SPI only.

jbouffard: Ah, I see what you're saying. Yeah, it's that way from now due to `GeoTrellisRasterSource` needing a `DataPath`. Though, now that I'm thinking about it, I'm not sure if that's really ideal :confused:

pomadchin: redundant in this case ): can you fix it?

jbouffard: Yeah, I can do that :+1:

vlm/src/test/scala/geotrellis/contrib/vlm/avro/GeotrellisRasterSourceSpec.scala
```diff
@@ -35,8 +35,8 @@ class GeotrellisRasterSourceSpec extends FunSpec with RasterMatchers with Better
   val uriMultibandNoParams = s"file://${TestCatalog.multibandOutputPath}"
   val uriMultiband = s"file://${TestCatalog.multibandOutputPath}?layer=${layerId.name}&zoom=${layerId.zoom}"
   val uriSingleband = s"file://${TestCatalog.singlebandOutputPath}?layer=${layerId.name}&zoom=${layerId.zoom}"
-  lazy val sourceMultiband = new GeotrellisRasterSource(uriMultiband, layerId)
-  lazy val sourceSingleband = new GeotrellisRasterSource(uriSingleband, layerId)
+  lazy val sourceMultiband = new GeotrellisRasterSource(uriMultiband)
```
pomadchin: Hm, you know; this looks better :+1:

vlm/src/main/scala/geotrellis/contrib/vlm/avro/GeoTrellisDataPath.scala
```diff
@@ -0,0 +1,90 @@
...
+ *      - '''band_count''': The number of bands of each Tile in the layer.
+ *    Of the above three parameters, `layer` and `zoom` are required. In addition,
+ *    this path can be prefixed with, '''gt+''' to signify that the target path
+ *    is to be read in only by [[GeotrellisRasterSource]].
+ *  @example "s3://bucket/catalog?layer=layer_name&zoom=10"
+ *  @example "hdfs://data-folder/catalog?layer=name&zoom-12&band_count=5"
+ *  @example "gt+file:///tmp/catalog?layer=name&zoom=5"
+ *  @note The order of the query parameters does not matter.
+ */
+case class GeoTrellisDataPath(path: String) extends DataPath {
```
pomadchin: Can you put into the companion object a helper method that will create a `GeoTrellisDataPath` from the `pathToCatalog` and `LayerId`?

PR #195, opened 2019-06-26
Use scala-uri library for URI parsing

### LINKED ISSUE #192: Use scala-uri library to parse VSIPath
Tight now we have lot's of weirdly implemented regexps, we can replace them by a lightweight [scala-uri](https://github.com/lemonlabsuk/scala-uri) library.

```scala
import io.lemonlabs.uri.{Url, UrlWithAuthority}

val fileName = "file-1.tiff"

val filePath = "www.radomdata.com/test-files/file-1.tiff"
val url = s"http://$filePath"
val httpUrl = UrlWithAuthority.parse(url)
httpUrl.schemeOption
httpUrl.hostOption
httpUrl.path
httpUrl.authority

val zipPath = "www.radomdata.com/test-files/data.zip"
val zipURL = s"zip+http://$zipPath"
val zipUrl = UrlWithAuthority.parse(zipURL)
zipUrl.schemeOption
zipUrl.hostOption
zipUrl.path
zipUrl.authority
```

It is also serializable, covered with tests, doesn't bring any really new deps (only a parbolied parser) and will help us to make the code more readable.

jbouffard:
# Overview

This PR improves the logic/readability of the path abstraction by using the `scala-uri` library.

## Checklist

- [ ] Add entry to CHANGELOG.md

Closes #192


gdal/src/test/scala/geotrellis/contrib/vlm/gdal/GDALDataPathSpec.scala
```diff
@@ -336,31 +281,24 @@ class GDALDataPathSpec extends FunSpec with Matchers {
...
-  }
 
-  describe("Formatting the given uris - edge cases") {
-    it("should parse a path with uncommon characters") {
-      val filePath = """data/jake__user--data!@#$%^&*()`~{}[]\|=+,?';<>;/files/my-data.tif"""
-      val uri = s"s3://$filePath"
-      val expectedPath = s"/vsis3/$filePath"
+    describe("Formatting the given uris - edge cases") {
+      ignore("should parse a path with uncommon characters") {
+        val filePath = """data/jake__user--data!@#$%^&*()`~{}[]\|=+,?';<>;/files/my-data.tif"""
```
jbouffard: Unfortunately, `scala-uri` isn't able to handle certain characters out of the box, and this test doesn't pass because of it. There's probably a way to change the encoding, but I'd have to look into this more.

pomadchin: @jbouffard double check how you use the library, here is some discussion: https://github.com/lemonlabsuk/scala-uri/issues/42

jbouffard:
[This test](https://github.com/jbouffard/geotrellis-contrib/blob/feature/path-abstraction-improvement/gdal/src/test/scala/geotrellis/contrib/vlm/gdal/GDALRasterSourceSpec.scala#L113) is currently failing because `GDALDataPath` has no way of throwing a `MalformedURLException`. This is because the `scala-uri` types are robust enough that they seem to be able to handle most bad `URI`s without issue. In addition, it's difficult to determine if a `URI` is bad within the logic of `GDALDataPath` and the `GDALPathType`s, as there are many different formats the path could be in.

It seems to me then, that the next best place to throw this exception would be when we're doing the actual reading. I'm not sure how I feel about this though, because it seems strange that a `MalformedURLException` would be thrown during reading and not when it was supposed to be formatted/constructed. However, i could just be over thinking it and that it's not really a big deal.

echeipesh:
@jbouffard Why do you expect the `MalformedURLException` in that case? `file:/random/path/here/N49W155.hgt.gz` reads as balid URL, what is malformed ?

jbouffard:
@echeipesh We'd expect one because that `URI` lacks an `authority`, which is currently a requirement.

pomadchin:
@jbouffard I think we can require `Uri` and match at runtime is it a Uri with authority or not.

echeipesh:
That URI would construct a Hadoop [Path](https://hadoop.apache.org/docs/stable/api/org/apache/hadoop/fs/Path.html) without an issue. Even in general `authority` is [optional](https://en.wikipedia.org/wiki/Uniform_Resource_Identifier) part of URI.  I'm a little bit out of the loop on this PR, where is it currently required?

pomadchin:
@echeipesh I think it is more about the scala-uri API now; dunno why it is required now, but at least Azure URIs require authority; but we don't use Azure `wasbs` uris now

jbouffard:
@echeipesh There are two types paths that can be passed to `GDALDataPath`: local, relative paths and `URI`s. If given a `URI`, it must have an `authority` mostly to keep the logic as simple as possible. I feel like if we don't put certain constraints on the format of the `URI`s it's going to be very difficult to parse/format them correctly.

@pomadchin I already do something similar where I check to see if it has an authority or not. Unfortunately, this doesn't quite work as if it doesn't have an authority, it's assumed to be a local path.

gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALDataPath.scala
```diff
@@ -43,9 +45,24 @@ case class GDALDataPath(
   path: String,
   compressedFileDelimiter: String = "!"
```
echeipesh: Is it ever the case that the compressed file delimiter changes or is it basically GDAL convention ?

jbouffard: I'm not sure if it's a GDAL convention or not, but I believe it is in Rasterio. The only time I can think of would be if that character is used in a file name somewhere.

gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALDataPath.scala
```diff
@@ -43,9 +42,26 @@ case class GDALDataPath(
...
+    else
+      Url.parseOption(path) match {
+        case Some(uri) => URIPath(uri)
+        case None =>
+          val encodedPath = UrlPath.fromRaw(path).toString
+          URIPath(Url.parse(encodedPath))
+      }
 
   private val badPrefixes: List[String] =
     List("gt+", "gtiff+")
```
echeipesh: This is a pretty general comment but I think its worth mentioning here. The idea that while parsing `GDAL` path we have a list of things that we know are bad goes against the general design principle of [separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns). 

What happens when more prefixes are added, does the logic in this class become invalid? And why would adding a new reading mechanism, that is unrelated to GDAL would require changing class concerned only with GDAL paths. These prefixes do serve a function to disambiguate URIs that can be parsed by multiple readers but such checking should happen outside of reader specific code and where the ambiguity exists, delegating the full parsing to the appropriate module/class.

pomadchin:
replaced by #201

PR #201, opened 2019-07-01
Use scala-uri library for URI parsing

### LINKED ISSUE #191: VSIPath inconsistent behavior
For the context, originally we were copying a [RasterIO behavior](https://github.com/mapbox/rasterio/blob/960a906dad2a4e426387ce048a52c6e90afdcd2b/tests/test_path.py).

The example of our tests are here [https://github.com/geotrellis/geotrellis-contrib/blob/master/gdal/src/test/scala/geotrellis/contrib/vlm/gdal/VSIPathSpec.scala#L59-L73]. 

The question is: why do we add `visgzip` prefix automatically, guessing about it from the file extension and we don't do the same way with `zip` extension: 

```scala
it("http that points to gz with ! url") {
  val filePath = "www.radomdata.com/test-files/data.gz"
  val url = s"http://$filePath!$fileName"
  val expectedPath = s"/vsigzip//vsicurl/http://$filePath/$fileName"

  VSIPath(url).vsiPath should be (expectedPath)
}

it("zip+http url") {
  val filePath = "www.radomdata.com/test-files/data.zip"
  val url = s"zip+http://$filePath"
  val expectedPath = s"/vsizip//vsicurl/$url"

  VSIPath(url).vsiPath should be (expectedPath)
}
```

It makes sense to choose one approach and follow it, to avoid such an inconsistent VSIParser behavior.

Edit: 

I also noticed this test case:

```scala
it("file that points to zip with ! uri") {
  val filePath = "/home/jake/Documents/test-files/files.zip"
  val uri = s"file://$filePath!$fileName"
  val expectedPath = s"/vsizip/$filePath/$fileName"

  VSIPath(uri).vsiPath should be (expectedPath)
}
```

It means, that `zip+` prefix is not necessary at all. Should we keep it then? What are your thoughts about the implicit `vsizip/vsigzip` application?

### LINKED ISSUE #192: Use scala-uri library to parse VSIPath

pomadchin:
# Overview

Replaces https://github.com/geotrellis/geotrellis-contrib/pull/195

## Checklist

Closes #192
Closes #191


gdal/src/test/scala/geotrellis/contrib/vlm/gdal/GDALDataPathSpec.scala
```diff
@@ -335,14 +345,22 @@ class GDALDataPathSpec extends FunSpec with Matchers {
...
-        GDALDataPath(uri).vsiPath should be (expectedPath)
-      }
+      GDALDataPath(expectedPath).vsiPath should be (expectedPath)
+    }
+
+    it("should parse a chained VSI path") {
+      val filePath = "/vsizip//vsis3/data/my-data/data.zip"
+      val expectedPath = s"/vsizip//vsis3/$filePath"
+
+      GDALDataPath(expectedPath).vsiPath should be (expectedPath)
```
echeipesh: These tests look odd, `expectedPath` would be `/vsizip//vsis3//vsizip//vsis3/data/my-data/data.zip` which isn't a thing

gdal/src/test/scala/geotrellis/contrib/vlm/gdal/GDALRasterSourceSpec.scala
```diff
@@ -109,8 +109,11 @@ class GDALRasterSourceSpec extends FunSpec with RasterMatchers with BetterRaster
       GDALRasterSource(GDALDataPath(p)).extent should be (GeoTiffRasterSource(p).extent)
     }
 
-    it("should fail on creation of the GDALRasterSource on a malformed URI") {
-      an[MalformedURLException] should be thrownBy GDALRasterSource(GDALDataPath("file:/random/path/here/N49W155.hgt.gz"))
+    ignore("should fail on creation of the GDALRasterSource on a malformed URI") {
+      //an[MalformedURLException] should be thrownBy GDALRasterSource(GDALDataPath("file:/random/path/here/N49W155.hgt.gz"))
+      val result = GDALRasterSource(GDALDataPath("file:/random/path/here/N49W155.hgt.gz"))
+      println(s"\n\nThis is the result: ${result.vsiPath}")
+      println(s"This is the result: ${Schemes.isVSIFormatted(result.vsiPath)}")
```
echeipesh: Why `ignore`? The path doesn't exist but it reads as valid, I think it should parse well?

gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALDataPath.scala
```diff
@@ -41,261 +42,75 @@ import java.net.MalformedURLException
  */
 case class GDALDataPath(
   path: String,
-  compressedFileDelimiter: String = "!"
+  compressedFileDelimiter: Option[String] = "!".some,
+  percentEncoder: PercentEncoder = PercentEncoder(PATH_CHARS_TO_ENCODE ++ Set('%', '?', '#'))
```
echeipesh: Are these truly parameters? It seems odd to have them here in the `case class` and also I'm not sure they ever change. C/D?

gdal/src/test/scala/geotrellis/contrib/vlm/gdal/GDALRasterSourceSpec.scala
pomadchin: It was copied from Jakes tests, dunno

gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALDataPath.scala
pomadchin: @echeipesh without these params some weird S3 paths may not work; actually I don't know what is the entire list of symbols we want to encode, that's why I decided to make it a parameter, that can be a parameter of the parsing function that would be in a companion object (we talked about it earlier today).

echeipesh: Naive question, should we just use [`URLDecoder`](https://docs.oracle.com/javase/7/docs/api/java/net/URLDecoder.html) because presumably it handles all the cases?

gdal/src/test/scala/geotrellis/contrib/vlm/gdal/GDALDataPathSpec.scala
pomadchin: These are incorrect tests back ported from the previous PR

gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALDataPath.scala
pomadchin: @echeipesh will give it a try

pomadchin: @echeipesh 

```
------PercentEncoder---------
/Users/me/subversions/git/github/locationtech/geotrellis-contrib/gdal/target/scala-2.11/test-classes/img/aspect-tiled.tif
------------------------------
%2FUsers%2Fme%2Fsubversions%2Fgit%2Fgithub%2Flocationtech%2Fgeotrellis-contrib%2Fgdal%2Ftarget%2Fscala-2.11%2Ftest-classes%2Fimg%2Faspect-tiled.tif
------URLEncoder--------------
```

gdal/src/test/scala/geotrellis/contrib/vlm/gdal/GDALRasterSourceProviderSpec.scala
```diff
@@ -38,7 +37,7 @@ class GDALRasterSourceProviderSpec extends FunSpec {
     }
 
     it("should not be able to process a GeoTrellis catalog path") {
-      assert(!provider.canProcess("s3://path/to/my/fav/catalog?layer=fav&zoom=3"))
+      assert(!provider.canProcess("gt+s3://path/to/my/fav/catalog?layer=fav&zoom=3"))
```
jbouffard: What happens if I give it `s3://path/to/my/fav/catalog?layer=fav&zoom=3`? Would it still know that it can't process it?

vlm/src/main/scala/geotrellis/contrib/vlm/avro/GeoTrellisDataPath.scala
```diff
@@ -37,54 +40,48 @@ import geotrellis.spark.LayerId
...
-      .split('&')
-      .map {
-        _.split('=') match {
-          case Array(k, v) => (k, v)
-        }
+  // try to parse it, otherwise it is a path
+  private val uri = UrlWithAuthority.parseOption(path).fold(Url().withPath(UrlPath.fromRaw(path)): Url)(identity)
+  private val queryString = uri.query
+
+  lazy val catalogPath: String = {
```
jbouffard: Does this `val` need to be `lazy`? I understand that `lazy val`s can give us trouble during run time.

vlm/src/main/scala/geotrellis/contrib/vlm/geotiff/GeoTiffDataPath.scala
```diff
@@ -30,50 +33,32 @@ import geotrellis.contrib.vlm.DataPath
...
-  private val servicePrefix: String = "gtiff+"
+  private val upath = percentEncoder.encode(path, "UTF-8")
 
   /** The formatted path to the GeoTiff that will be read */
-  def geoTiffPath: String =
-    if (path.startsWith(servicePrefix))
-      path.splitAt(servicePrefix.size)._2
-    else
-      path
+  lazy val geoTiffPath: String = {
```
jbouffard: Same question as above.

vlm/src/main/scala/geotrellis/contrib/vlm/geotiff/GeoTiffDataPath.scala
```diff
@@ -30,50 +33,32 @@ import geotrellis.contrib.vlm.DataPath
...
+    // this parser supports only TIF and .tif exts
```
jbouffard: Why? Why not all the user to add additional extensions to look for?

gdal/src/test/scala/geotrellis/contrib/vlm/gdal/GDALRasterSourceSpec.scala
pomadchin: @echeipesh there is no scheme in this uri; parser wont figure out that file:/ is a typo and it should be file://

gdal/src/test/scala/geotrellis/contrib/vlm/gdal/GDALRasterSourceProviderSpec.scala
pomadchin: @jbouffard no, why GDAL can't read smth from `s3://path/to/my/fav/catalog?layer=fav&zoom=3`? In theory this can be some valid path with smth on that side being understand by GDAL. What is your suggestion? You think `http://domain/to/my/fav/catalog?layer=fav&zoom=3` should not work as well?

vlm/src/main/scala/geotrellis/contrib/vlm/avro/GeoTrellisDataPath.scala
pomadchin: I thought about moving the parsing logic our of the `GeoTrellisDataPath` object.

vlm/src/main/scala/geotrellis/contrib/vlm/geotiff/GeoTiffDataPath.scala
pomadchin: @jbouffard copied your mistaken logic, need to remove it indeed and remove the test.

gdal/src/test/scala/geotrellis/contrib/vlm/gdal/GDALRasterSourceProviderSpec.scala
jbouffard: I'm just wondering if the SPI would pick up on that. Since wouldn't both `GDAL` and `GeoTrellis` `RasterSource` think they can process it?

pomadchin: I think it is impossible to dispatch in general (?); the right way is always to require prefix for GeoTrellis avro layers, since GDAL can work with anything that has an array of bytes with proper headers on the other side.

gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALRasterSourceProvider.scala
```diff
@@ -17,17 +17,12 @@
...
 
 class GDALRasterSourceProvider extends RasterSourceProvider {
   def canProcess(path: String): Boolean =
-    try {
-      GDALDataPath(path)
-      true
-    } catch {
-      case _: Throwable => false
-    }
+    (!path.startsWith(GeoTiffDataPath.PREFIX) && !path.startsWith(GeoTrellisDataPath.PREFIX)) && GDALDataPath.parseOption(path).nonEmpty && path.nonEmpty
```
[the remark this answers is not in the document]
pomadchin: I thought that we've talked about it during paring, I think this logic `gt+`, `gdal+` and `gtiff+` it is only an SPI logic. Our parsers are agnostic to these prefix. 

Also `RasterSource(gtPath)` can return the actuall tiff, in case on the return side there would be a properly encoded GeoTiff with proper headers (yes, GDAL can read things that have weird URIs and the data extension as well as the data uri can be of different formats).

`gt+http://path/to/my/catalog?layer=name&zoom=10` can be also parsed by the `GDALRasterSource` properly.

gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALDataPath.scala
```diff
@@ -35,267 +36,90 @@ import java.net.MalformedURLException
...
-
-  implicit def toGDALDataPath(path: String): GDALDataPath =
-    GDALDataPath(path)
+  def parse(
+    path: String,
+    compressedFileDelimiter: Option[String] = "!".some,
+    percentEncoder: PercentEncoder = PercentEncoder(PATH_CHARS_TO_ENCODE ++ Set('%', '?', '#'))
+  ): GDALDataPath =
+    parseOption(path, compressedFileDelimiter, percentEncoder)
+      .getOrElse(throw new IllegalArgumentException(s"Unable to parse GDALDataPath: $path"))
```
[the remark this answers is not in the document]
pomadchin: @jbouffard it can be also a local path, in this case it won't be a URL, right?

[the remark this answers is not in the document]
jbouffard: Ah, I guess that's true.

gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALRasterSourceProvider.scala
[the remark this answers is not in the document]
jbouffard: Yeah, you're right. Now that I'm thinking about it, we did discuss it. That makes sense :+1:

gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALDataPath.scala
```diff
@@ -35,267 +36,90 @@ import java.net.MalformedURLException
...
+
+              // check out the last .${extension}, probably we need auto add it into the vsipath construction
+              val extraScheme =
+                COMPRESSED_FILE_TYPES
+                  .flatMap { ext => if (path.contains(s".$ext")) toVSIScheme(ext).some else None }
+                  .lastOption
+
+              // check out that we won't append a vsi path duplicate or other compression vsipath
+              val extraSchemeExists = extraScheme.exists { es =>
+                schemes.nonEmpty && (schemes.contains(es) || COMPRESSED_FILE_TYPES.map(toVSIScheme).collect { case es if es.nonEmpty => schemes.contains(es) }.reduce(_ || _))
```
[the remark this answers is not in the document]
pomadchin: ¯\_(ツ)_/¯ I don't find that your way of writing the condition is more readable even with a better name like `val alreadyCompressed`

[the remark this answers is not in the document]
pomadchin: Hm, mb we can a function that will incapsulate this logic and will make a more readable application.

pomadchin:
Ok @moradology 2 > 1 I'm buying it :+1:

Issue #102, opened 2019-01-17
RasterSource.reproject to CRS of underlying file should short-circuit


echeipesh:
Sometimes reprojection is done carelessly or reprojection of reprojected raster source could take it back to the starting CRS. In the cases we should certainly avoid returning `_ReprojectedRasterSource` and instead try to return either the base or the resampled version.

This is not a critical issue but important for performance and to minimize possible distortions of going through needless reprojection.

echeipesh:
`GeoTrellisRasterSource` logic has been updated to do this in as of `3.11`

pomadchin:
https://github.com/geotrellis/geotrellis-contrib/pull/202 also aligns with the idea of this PR to short-circuit the transformations; what needs to be done in particular (one of the possible solutions implementation strategy): 

1. make the [current reprojection method protected and rename it into smth internal](https://github.com/geotrellis/geotrellis-contrib/blob/aca902c5c9ff8e37169b6e8f95bc22decf7ae5e2/vlm/src/main/scala/geotrellis/contrib/vlm/RasterSource.scala#L87)
2. implement a new one (with the name of the current) that will introduce the logic of the CRS checks

This will allow us to separate the pure reprojection logic from the CRS checks and will make the methods behavior more transparent. 

These changes are definitely required for the proper `GeoTiffRasterSources` work and for the `GDALRasterSources` as well (since in GDALRasterSources we're [stacking sequences of GDALWarpOptions transformations](https://github.com/geotrellis/geotrellis-contrib/blob/3d05e9c9ae45c68266b3c2f6a556a303102c0115/gdal/src/main/scala/geotrellis/contrib/vlm/gdal/Implicits.scala#L54) into a single GDALWarp object). 

The implementation of these changes can be a bit different from what I described i.e. without having a protected method just introduce an extra CRS check directly into the reproject function implementation, however it won't make the code more readable.

PR #202, opened 2019-07-01
Fix Incomplete Reads for MosaicRasterSource

notthatbreezy:
# Overview

MosaicRasterSource does not return a result with an extent that matches the requested extent if the requested extent is outside the bounds for the mosaic raster source.

This is a WIP while the fix is still being worked on, but the first commit adds a failing test case.

## Testing Instructions

I added a test that reuses some of the existing tiffs in `MosaicRasterSourceSpec` -- that should pass. I don't think there are any additional tests required besides that.

## Checklist

Closes https://github.com/geotrellis/geotrellis-server/issues/124


vlm/src/main/scala/geotrellis/contrib/vlm/MosaicRasterSource.scala
```diff
@@ -99,7 +100,9 @@ trait MosaicRasterSource extends RasterSource {
   )
 
   def read(extent: Extent, bands: Seq[Int]): Option[Raster[MultibandTile]] = {
-    val rasters = sources map { _.read(extent, bands) }
+    val rasters = sources map {
+      _.read(extent, bands).map(_.crop(extent, Crop.Options(false, false)))
```
pomadchin: We've talked with @echeipesh and @moradology, and discovered that there is a problem in the constructor: that MosaicRasterSource constructor resamples to a potentially incorrect grid in the incorrect CRS the original rasterSource: https://github.com/geotrellis/geotrellis-contrib/blob/master/vlm/src/main/scala/geotrellis/contrib/vlm/MosaicRasterSource.scala#L156

So this fix is not a fix and breaks all the logic.

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSource.scala
```diff
@@ -102,7 +102,12 @@ trait RasterSource extends CellGrid[Long] with Serializable {
     * @group reproject a
     */
   def reprojectToGrid(crs: CRS, grid: GridExtent[Long], method: ResampleMethod = NearestNeighbor, strategy: OverviewStrategy = AutoHigherResolution): RasterSource =
-    reproject(crs, Reproject.Options(method = method, parentGridExtent = Some(grid)), strategy)
+    if (crs == this.crs && grid == this.gridExtent)
+      this
+    else if (crs == this.crs)
+      resampleToGrid(grid)
```
moradology: I'm unclear on whether doing the 'smart' thing here is actually desirable. It might make the API more magical

pomadchin: i'm still looking into this PR; however, answering your comment; I think in general idea - it totally fine aligns with this issue https://github.com/geotrellis/geotrellis-contrib/issues/102

Issue #208, opened 2019-07-10
Bump Proj4 dependency version up


pomadchin:
Right now we [have to lock on the reprojection](https://github.com/geotrellis/geotrellis-contrib/pull/206/files#diff-7e0054c3d5e2a11ee0e42f22f4b06d8fR102) since `BasicCoordinateTransform` object [is not thread-safe](https://github.com/locationtech/proj4j/pull/29). This issue is a place holder to keep in mind and to remove this lock once a new proj4 version would be released.

pomadchin:
It also would address https://github.com/geotrellis/geotrellis-contrib/issues/226

PR #209, opened 2019-07-11
Reprojection Short Circuting

### LINKED ISSUE #102: RasterSource.reproject to CRS of underlying file should short-circuit

jbouffard:
# Overview

This PR adds logic that checks to see if the target `RasterSource` need to be reprojected. If they don't, then they are either resampled are just returned as-is.

## Checklist

Closes #102

### DIFF
--- a/vlm/src/main/scala/geotrellis/contrib/vlm/RasterSource.scala
+++ b/vlm/src/main/scala/geotrellis/contrib/vlm/RasterSource.scala
@@ -80,11 +80,15 @@ trait RasterSource extends CellGrid[Long] with Serializable {
   /** Raster pixel row count */
   def rows: Long = gridExtent.rows
 
+  protected def reprojection(crs: CRS, reprojectOptions: Reproject.Options, strategy: OverviewStrategy): RasterSource
+
   /** Reproject to different CRS with explicit sampling reprojectOptions.
     * @see [[geotrellis.raster.reproject.Reproject]]
     * @group reproject
     */
-  def reproject(crs: CRS, reprojectOptions: Reproject.Options, strategy: OverviewStrategy): RasterSource
+  def reproject(crs: CRS, reprojectOptions: Reproject.Options, strategy: OverviewStrategy): RasterSource =
+    if (crs == this.crs) this
+    else reprojection(crs, reprojectOptions, strategy)
 
   /** Sampling grid is defined over the footprint of the data at default resolution
     *
@@ -94,7 +98,8 @@ trait RasterSource extends CellGrid[Long] with Serializable {
     * @group reproject
     */
   def reproject(crs: CRS, method: ResampleMethod = NearestNeighbor, strategy: OverviewStrategy = AutoHigherResolution): RasterSource =
-    reproject(crs, Reproject.Options(method = method), strategy)
+    if (crs == this.crs) this
+    else reprojection(crs, Reproject.Options(method = method), strategy)
 
   /** Sampling grid and resolution is defined by given [[GridExtent]].
     * Resulting extent is the extent of the minimum enclosing pixel region
@@ -103,8 +108,8 @@ trait RasterSource extends CellGrid[Long] with Serializable {
     */
   def reprojectToGrid(crs: CRS, grid: GridExtent[Long], method: ResampleMethod = NearestNeighbor, strategy: OverviewStrategy = AutoHigherResolution): RasterSource =
     if (crs == this.crs && grid == this.gridExtent) this
-    else if (crs == this.crs) resampleToGrid(grid)
-    else reproject(crs, Reproject.Options(method = method, parentGridExtent = Some(grid)), strategy)
+    else if (crs == this.crs) resampleToGrid(grid, method)
+    else reprojection(crs, Reproject.Options(method = method, parentGridExtent = Some(grid)), strategy)
 
   /** Sampling grid and resolution is defined by given [[RasterExtent]] region.
     * The extent of the result is also taken from given [[RasterExtent]],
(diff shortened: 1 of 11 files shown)


vlm/src/main/scala/geotrellis/contrib/vlm/RasterSource.scala
```diff
@@ -94,7 +98,8 @@ trait RasterSource extends CellGrid[Long] with Serializable {
     * @group reproject
     */
   def reproject(crs: CRS, method: ResampleMethod = NearestNeighbor, strategy: OverviewStrategy = AutoHigherResolution): RasterSource =
-    reproject(crs, Reproject.Options(method = method), strategy)
+    if (crs == this.crs) this
+    else reproject(crs, Reproject.Options(method = method), strategy)
```
echeipesh: The PR does not pass the CI yet, but when it does I'm pretty sure this is going to be a circular reference that blows the stack.

Thinking about this I think there are two things that make me feel weird about the protected approach

1. `_reproject` - the underscore convention universally means "private variable, don't pay attention to this"  so it's weird to be asking people to implement it.
2. `_reproject` - name looks pretty similar to `reproject` that can easily lead to careless mistake like the one here thus adding a little bit of misery to API user.

The only name I can think of is `reprojectInternal` which ... is different but not sure is super meaningful.
++ @pomadchin for naming discussion
++ @metasim as our start external user for this API

pomadchin: I agree that `_reproject` is a bad name. Probably it is because of me it appeared in this PR ([since I mentioned it for some reason here](https://github.com/geotrellis/geotrellis-contrib/issues/102#issuecomment-509371493) just to give a better task description).

Is there an `underscore convention` in scala? As I it is said [here](https://docs.scala-lang.org/style/naming-conventions.html) it looks like it is a convention for `mutators`.

Names that come to my mind are `reprojectBase`, `reprojectRaw`, `reprojectUnsafe`, `reprojectForce`

jbouffard: I agree that `_reproject` isn't really the best name, but I was unable to come up with anything better. I also don't think the `under score convention` is supposed to be used this way in Scala. However, I noticed that we use this style in some places in GeoTrellis

https://github.com/locationtech/geotrellis/blob/master/raster/src/main/scala/geotrellis/raster/io/geotiff/GeoTiffMultibandTile.scala#L811

https://github.com/locationtech/geotrellis/blob/master/spark/src/main/scala/geotrellis/spark/store/LayerWriter.scala#L124

pomadchin: In case of a `GeoTiffMultibandTile.scala` it is okay (IMO) since nobody would ever use it / overload it in the client API.

In case of LayerWriter it is an oversight.

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSource.scala
```diff
@@ -80,11 +80,15 @@ trait RasterSource extends CellGrid[Long] with Serializable {
   /** Raster pixel row count */
   def rows: Long = gridExtent.rows
 
+  protected def _reproject(crs: CRS, reprojectOptions: Reproject.Options, strate: OverviewStrategy): RasterSource
```
moradology: `strate` -> `strategy`

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSource.scala
```diff
@@ -80,11 +80,15 @@ trait RasterSource extends CellGrid[Long] with Serializable {
...
+
   /** Reproject to different CRS with explicit sampling reprojectOptions.
     * @see [[geotrellis.raster.reproject.Reproject]]
     * @group reproject
     */
-  def reproject(crs: CRS, reprojectOptions: Reproject.Options, strategy: OverviewStrategy): RasterSource
+  def reproject(crs: CRS, reprojectOptions: Reproject.Options, strategy: OverviewStrategy): RasterSource =
+    if (crs == this.crs) this
```
metasim: Be aware that this can be very computationally expensive. See https://github.com/locationtech/geotrellis/issues/3022

metasim: See also: https://github.com/locationtech/rasterframes/issues/134

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSource.scala
moradology: This PR looks good but I'm not really comfortable with something being called `internal` which is actually required of any would-be implementers. We can either find a better name (struggling to do that, personally) or expect implementers to provide their own short circuiting logic within `reproject`

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSource.scala
pomadchin: @metasim https://github.com/geotrellis/geotrellis-contrib/issues/208

moradology:
I've thought about this a bit and I've got a potential solution which may be controversial. Instead of 'reproject' being the externally facing method, we could find an idempotent way of phrasing things (short circuiting ensures idempotence). Thus, we might have:
```scala
val rs: RasterSource = GeotiffRasterSource(...)
rs.projected(crs, reprojectOptions, strategy)
```

Our expectation via the trait, on the other hand, would be `reproject` which only gets called if the projection isn't already the one requested

jbouffard:
@moradology @echeipesh @pomadchin have we decided on a naming strategy for the externally facing method?

moradology:
Another option brought up by @pomadchin which I think deserves our attention and would be simpler for now is to use `reproject` as we currently are and change `reprojectInternal` to `reprojection`

jbouffard:
I think they're both good options. @pomadchin's might be better in the short term, as it doesn't require a refactor, but I think we should implement something similar to what @moradology suggested later on.

echeipesh:
@jbouffard I think thats a reasonable compromise to close out this PR.

@jbouffard / @moradology 
Given that the iron is hot the issue for the the idempotent reprojection should include the mockup of effected methods and related. I continue to like the idea and I think it is worth further thought.

PR #210, opened 2019-07-12
GeoTiffRasterSource Tiff Reuse

### LINKED ISSUE #155: GeoTiffRasterSource should re-use base tiff
This issue is for possible optimization.

Currently `GeoTiffReprojectRasterSource` and `GeoTiffResampleRasterSource` does not accept base `GeoTiff` object:

https://github.com/geotrellis/geotrellis-contrib/blob/e0fc7d6b4d3cbe4d10c16c304a91a31b12a18d1d/vlm/src/main/scala/geotrellis/contrib/vlm/geotiff/GeoTiffReprojectRasterSource.scala#L28-L34

Instead they will re-open tiff https://github.com/geotrellis/geotrellis-contrib/blob/e0fc7d6b4d3cbe4d10c16c304a91a31b12a18d1d/vlm/src/main/scala/geotrellis/contrib/vlm/geotiff/GeoTiffReprojectRasterSource.scala#L37-L38

This will result in double fetch of metadata in the normal workflow like this:

```scala
val rs = GeoTiffRasterSource("s3://bucket/file.tiff")
val wm = rs.reprojectToGrid(WebMercator, layout)
wm.read(extent)
```
`rs.reprojectToGrid` will allocate `GeoTiffReprojectRasterSource` which will re-read the metadata which may have already been ready by `GeoTiffRasterSource`.

In contrast in process of optimizing `GeoTrellisRasterSource` passing already read layer metadata avoided this problem and noticeably helped performance:

https://github.com/geotrellis/geotrellis-contrib/blob/master/vlm/src/main/scala/geotrellis/contrib/vlm/avro/GeotrellisReprojectRasterSource.scala#L31-L40

Some things to consider:

- I'm not certain of what behavior is of `GeoTiff` class under multi-threaded read.
- We have not made any statements about thread safety of resource uniqueness of base an reprojected raster source, sounds like we might have to.
- I don't think there is a good way to separate the GeoTiff metadata from GeoTiff object

jbouffard:
# Overview

This PR improves the performance of `GeoTiffRasterSource`s by having the base tiff be an optional constructor. This will prevent the same Tiff being reread multiple times during a job.

## Checklist

Closes #155

### DIFF
--- a/vlm/src/main/scala/geotrellis/contrib/vlm/geotiff/GeoTiffReprojectRasterSource.scala
+++ b/vlm/src/main/scala/geotrellis/contrib/vlm/geotiff/GeoTiffReprojectRasterSource.scala
@@ -31,7 +31,7 @@ case class GeoTiffReprojectRasterSource(
   reprojectOptions: Reproject.Options = Reproject.Options.DEFAULT,
   strategy: OverviewStrategy = AutoHigherResolution,
   private[vlm] val targetCellType: Option[TargetCellType] = None,
-  val baseTiff: Option[MultibandGeoTiff] = None
+  private val baseTiff: Option[MultibandGeoTiff] = None
 ) extends RasterSource { self =>
   def resampleMethod: Option[ResampleMethod] = Some(reprojectOptions.method)
 
--- a/vlm/src/main/scala/geotrellis/contrib/vlm/geotiff/GeoTiffReprojectRasterSource.scala
+++ b/vlm/src/main/scala/geotrellis/contrib/vlm/geotiff/GeoTiffReprojectRasterSource.scala
@@ -30,12 +30,13 @@ case class GeoTiffReprojectRasterSource(
   crs: CRS,
   reprojectOptions: Reproject.Options = Reproject.Options.DEFAULT,
   strategy: OverviewStrategy = AutoHigherResolution,
-  private[vlm] val targetCellType: Option[TargetCellType] = None
+  private[vlm] val targetCellType: Option[TargetCellType] = None,
+  val baseTiff: Option[MultibandGeoTiff] = None
 ) extends RasterSource { self =>
   def resampleMethod: Option[ResampleMethod] = Some(reprojectOptions.method)
 
   @transient lazy val tiff: MultibandGeoTiff =
-    GeoTiffReader.readMultiband(getByteReader(dataPath.path), streaming = true)
+    baseTiff.getOrElse(GeoTiffReader.readMultiband(getByteReader(dataPath.path), streaming = true))
 
   protected lazy val baseCRS: CRS = tiff.crs
   protected lazy val baseGridExtent: GridExtent[Long] = tiff.rasterExtent.toGridType[Long]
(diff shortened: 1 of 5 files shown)


vlm/src/main/scala/geotrellis/contrib/vlm/geotiff/GeoTiffResampleRasterSource.scala
```diff
@@ -30,12 +30,13 @@ case class GeoTiffResampleRasterSource(
   resampleGrid: ResampleGrid[Long],
   method: ResampleMethod = NearestNeighbor,
   strategy: OverviewStrategy = AutoHigherResolution,
-  private[vlm] val targetCellType: Option[TargetCellType] = None
+  private[vlm] val targetCellType: Option[TargetCellType] = None,
+  private val baseTiff: Option[MultibandGeoTiff] = None
```
moradology: I wonder if this is actually something we need for all `GeoTiff` constructors - it seems to me like only the base `RasterSource` should have/needs this cached tiff

jbouffard: That wouldn't work as the base `RasterSource` doesn't contain any information regarding the format of the data (ie GeoTiffs, GDALDatasets, etc), but rather, just describes it; so we wouldn't be able to put caching logic there.

moradology: I'm not sure I follow. It appears to me that no extra logic is being done to modify `baseTiff`. My suggestion would be to simply remove `baseTiff` from all constructors other than the base (the `GeoTiffRasterSource` in this case)

jbouffard: Oh, I misunderstood. I thought you were referring to the `RasterSource` trait.

The problem with that is neither `GeoTiffReprojectRasterSoure` nor `GeoTiffResampleRasterSource` extends `GeoTiffRasterSource`. They're each a separate from one another. Also, there can be instances where the user creates either a `GeoTiffReprojectRasterSoure` or a `GeoTiffResampleRasterSource` directly, instead of first making a `GeoTiffRasterSource` and then doing some kind of transformation.

moradology: Misunderstandings all around... I incorrectly thought (this may have changed recently) that these `RasterSource` operations still involved encoding transformations via case class nesting.

LGTM in that case

vlm/src/main/scala/geotrellis/contrib/vlm/geotiff/GeoTiffReprojectRasterSource.scala
```diff
@@ -30,12 +30,13 @@ case class GeoTiffReprojectRasterSource(
   crs: CRS,
   reprojectOptions: Reproject.Options = Reproject.Options.DEFAULT,
   strategy: OverviewStrategy = AutoHigherResolution,
-  private[vlm] val targetCellType: Option[TargetCellType] = None
+  private[vlm] val targetCellType: Option[TargetCellType] = None,
+  val baseTiff: Option[MultibandGeoTiff] = None
```
pomadchin: Why is it `private val` in `GeoTiffRasterSource` and here it is `public`? What is the difference?

vlm/src/test/scala/geotrellis/contrib/vlm/geotiff/GeoTiffRasterSourceMultiThreadingSpec.scala
```diff
@@ -32,7 +32,7 @@ class GeoTiffRasterSourceMultiThreadingSpec extends AsyncFunSpec with Matchers {
   val source = GeoTiffRasterSource(url)
 
   implicit val ec = ExecutionContext.global
-  
+
```
pomadchin: It is an interesting change :D

vlm/src/main/scala/geotrellis/contrib/vlm/geotiff/GeoTiffReprojectRasterSource.scala
jbouffard: Oh, that's not right. That should be `private`.

vlm/src/test/scala/geotrellis/contrib/vlm/geotiff/GeoTiffRasterSourceMultiThreadingSpec.scala
jbouffard: I'm glad you think so haha

pomadchin:
LGTM, it loos like there is not too much we can do in GeoTiffRasterSources API to reuse the base tiff.

PR #211, opened 2019-07-15
GDALDataset Error Reporting

### LINKED ISSUE #179: Improve error reporting in GDALDataset
Reporting of errors from the JNI bindings could easily be improved. Currently they take the form of:

https://github.com/geotrellis/geotrellis-contrib/blob/94ccdebff8ce58f5f3d8e1f96046961bb6ba1a46/gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALDataset.scala#L148

 in the following ways:

1. Don't throw top-level `Exception`, but something more "filterable", such as `IOException`, `ConnectException` or a new custom `GDALException`.
2. Setting the message to be the name of the JNI method doesn't provide enough information to figure out what happened. First order would be to include the integer error code.
3. Even better would be to convert the error code via a lookup table into something user interpretable. e.g. "403 authorization error".

### LINKED ISSUE #215: Expose GDAL Error codes
It is a logical extension of the https://github.com/geotrellis/geotrellis-contrib/issues/179 

Exceptions that were added by closing https://github.com/geotrellis/geotrellis-contrib/issues/179 should also include error codes. By the time of creating this issue GDAL Warp bindings doesn't expose the underlying GDAL message and / or code: 

Example by @jbouffard:

```scala
scala> GDALWarp.get_band_count(token, GDALWarp.WARPED, numberOfAttempts, count)
ERROR 4: ../../blah.tif: No such file or directory
res5: Int = -1
```

Reference by @metasim about what do we mean here by [error codes](https://joinup.ec.europa.eu/svn/gvsig-3d/1.10/trunk/binaries/mac/raster/gdal/GDAL.framework/Versions/1.7/Resources/doc/gdal/cpl__error_8h.html). 

A related GDALWarpBindings issue: https://github.com/geotrellis/gdal-warp-bindings/issues/52

jbouffard:
# Overview

This PR changes the exceptions thrown in `GDALDataset` to be instances of a new `Exception` class `GDALException`. These new exceptions will provide more helpful information as to why a certain operation failed when reading in data with GDAL. The new class `GDALException` was created to allow users to more easily filter exceptions that are and are not GDAL related.

## Notes

This PR does not return error codes as there is currently no way to determine which one was thrown on the GDAL side. However, once https://github.com/geotrellis/gdal-warp-bindings/issues/52 is resolved, we'll be able to supply error codes along with the message.

## Checklist

Closes #179 
Closes #215


pomadchin:
Would also like to hear @metasim opinion on that, since he is the initial author of that and user of our VLM API

gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALDataset.scala
```diff
@@ -42,7 +42,7 @@ case class GDALDataset(token: Long) extends AnyVal {
     val width_height = Array.ofDim[Int](2)
     if (GDALWarp.get_transform(token, dataset, numberOfAttempts, transform) <= 0 ||
       GDALWarp.get_width_height(token, dataset, numberOfAttempts, width_height) <= 0)
-      throw new Exception("get_transform or get_width_height")
+      throw new MalformedDataException("Bad tranform values read")
```
pomadchin: Can you think of a better exception messages? In this case it looks like it can't construct `RasterExtent`.

gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALDataset.scala
```diff
@@ -68,7 +68,7 @@ case class GDALDataset(token: Long) extends AnyVal {
     val extent = this.extent(dataset)
 
     if (GDALWarp.get_overview_widths_heights(token, dataset, numberOfAttempts, 1, widths, heights) <= 0)
-      throw new Exception("get_overview_widths_heights")
+      throw new MalformedDataException("Unable to read in overviews")
```
pomadchin: It can not construct overivew `RasterExtents`. Mb it is better to throw an exception that would explain the issue cause in terms of the entire function definition?

gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALDataset.scala
```diff
@@ -145,7 +145,7 @@ case class GDALDataset(token: Long) extends AnyVal {
       val minmax = Array.ofDim[Double](2)
       val success = Array.ofDim[Int](1)
       if (GDALWarp.get_band_min_max(token, dataset, numberOfAttempts, 1, true, minmax, success) <= 0)
-        throw new Exception("get_band_min_max")
+        throw new MalformedDataTypeException("Bad min/max values")
```
pomadchin: What is the difference between `unable` and `bad` words you use in different exceptions?

gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALException.scala
```diff
@@ -0,0 +1,33 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package geotrellis.contrib.vlm.gdal
+
+
+/** Base class for all Exceptions involving GDAL. */
+class GDALException(message: String) extends Exception(message)
```
pomadchin: Should it be a `RuntimeException` like in `geotrellis-gdal`? I don't have a strong opinion about that btw.

gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALDataset.scala
```diff
@@ -42,7 +42,7 @@ case class GDALDataset(token: Long) extends AnyVal {
     val width_height = Array.ofDim[Int](2)
     if (GDALWarp.get_transform(token, dataset, numberOfAttempts, transform) <= 0 ||
       GDALWarp.get_width_height(token, dataset, numberOfAttempts, width_height) <= 0)
```
metasim: What are the possible values and semantics of the value returned integer value? If it can return values other than zero and one, I'd propose having the code included in the exception message so we can at least go look it up in the GDAL docs. We were getting errors thrown here that were due to S3 authentication errors, but at first it looked like there was a data error. Error codes could help ameliorate this.

gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALException.scala
metasim: @pomadchin `RuntimeException` should only be used for _unchecked_ exceptions (in a Java sense). While all exceptions in Scala are unchecked, I'd suggest following the general intent, and keeping `Exception` as the base type.

> RuntimeException is the superclass of those exceptions that can be thrown during the normal operation of the Java Virtual Machine.
> RuntimeException and its subclasses are unchecked exceptions. Unchecked exceptions do not need to be declared in a method or constructor's throws clause if they can be thrown by the execution of the method or constructor and propagate outside the method or constructor boundary.

gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALException.scala
```diff
@@ -0,0 +1,33 @@
...
+
+/** Exception thrown when data coulld not be read from data source. */
+class GDALIOException(message: String) extends Exception(message)
```
metasim: I suggest making this a subtype of `IOException`. 

Also: "coulld".

gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALException.scala
metasim: That said.... I could see debates on both side of the issue, so I wouldn't fight over my position ;-)

gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALDataset.scala
metasim: Yeh, according to [this](https://github.com/geotrellis/gdal-warp-bindings/blob/58d496521695b4bae2b7f7ab7d1f17140be0a569/src/bindings.cpp#L511-L512) the return value has some sort of meaning one could inspect for debugging.

gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALException.scala
metasim: Another  consideration, assuming all these exceptions are raised when getting error codes from the underlying GDAL library, is to have a field called `gdal_error_code` to capture the value returned.

pomadchin: @metasim yea in Scala all exceptions are unchecked, but I was just curious does it make too much sense to follow Java logic or not (like MalformedArgument exceptions, etc). But I see your comment about IO and I'm good with keeping it `<: Exception`

gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALDataset.scala
jbouffard: Really the only difference is that the `Unable` implies that something is `bad`. However, I think it'd be better if we tried to standardized the messages more, so I'll change them to one phrasing.

gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALDataset.scala
pomadchin: @metasim I think it returns a number of the datasets that `GDALWarp.method` touched to return a value requested from by the `method`. As @jamesmcclain mentioned right now, that the idea was that probably it would be possible to use this information in some dynamic performance tuning. There is an issue https://github.com/geotrellis/gdal-warp-bindings/issues/52 about it.

pomadchin: wooops @metasim mb you're right; @jbouffard can you check that it returns GDAL error codes? As it can return smth internal?

jbouffard: @pomadchin Do you have any malformed GeoTiffs? I don't think I have anything that would fail to be read

pomadchin: @jbouffard you can point to a non existent file or to a file that is not a geotiff though have a `.tiff` extension

jbouffard: @pomadchin @metasim This is what's returned when trying to read a bad file with GDAL:

```
scala> GDALWarp.get_band_count(token, GDALWarp.WARPED, numberOfAttempts, count)
ERROR 4: ../../blah.tif: No such file or directory
res5: Int = -1
```

So while he `ERROR` message shown is correct, the return of `-1` is not.

pomadchin: @jbouffard it looks like the only thing we can do in this case is to return to this problem, once it would be addressed in bindings. The best thing we can do now is to write our messages without codes, since it is impossible to retrieve that information from the native interface now.

gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALException.scala
jbouffard: This is something that should be added once the return value reflects the number of the error thrown.

gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALDataset.scala
```diff
@@ -68,7 +68,7 @@ case class GDALDataset(token: Long) extends AnyVal {
     val extent = this.extent(dataset)
 
     if (GDALWarp.get_overview_widths_heights(token, dataset, numberOfAttempts, 1, widths, heights) <= 0)
-      throw new Exception("get_overview_widths_heights")
+      throw new MalformedDataException("Unable to construct overviews for resample")
```
pomadchin: just a minor thing, mb contruct `overviews RasterExtents` (pr smth in a better language)? Since here we don't construct overviews, just overviews `RasterExtents`.

pomadchin:
Also created a follow up issue https://github.com/geotrellis/geotrellis-contrib/issues/215 that would be a separate small once, since it can not be done at the moment in terms of this PR.

jamesmcclain:
In progress https://github.com/geotrellis/gdal-warp-bindings/pull/56

pomadchin:
@jbouffard let's wait for a little bit? mb @jamesmcclain will release a new bindings version soon

jbouffard:
@pomadchin I was thinking that as well. No sense in creating another PR right after this one.

pomadchin:
@jbouffard James published already a new release! https://github.com/geotrellis/gdal-warp-bindings/pull/56

gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALException.scala
pomadchin: @jbouffard I think this comment is still unaddressed.

gdal/src/test/scala/geotrellis/contrib/vlm/gdal/GDALErrorSpec.scala
```diff
@@ -0,0 +1,32 @@
...
+
+  describe("GDALErrors") {
+    it("should return the ObjectNull error code") {
+      val result = GDALWarp.get_data(token, dataset, 1, sourceWindow, destWindow, 42, 1, buffer)
+
+      assert(math.abs(result) == 10)
+    }
+
+    it("should return the IllegalArg error code") {
+      val result = GDALWarp.get_data(token, dataset, 1, sourceWindow, destWindow, 1, 1, buffer)
```
pomadchin: Nice that it works! The only thing that needs to be done is to pass these error codes into our Exception types.

PR #212, opened 2019-07-15
Lazy Vals Marked as Transient

### LINKED ISSUE #82: Should mark lazy values as @transient(?)
Caching a `GeotiffRasterSource` causes NPE when the lazy vals are accessed a second time (unless the val is called and laziness is voided prior to caching). Marking lazy vals as transient might prevent this issue

jbouffard:
# Overview

This PR marks all `lazy val`s in the `RasterSource`s as `@transient` to help prevent NPEs from occurring in distributed work.

## Checklist

Closes #82

### DIFF
--- a/gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALRasterSource.scala
+++ b/gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALRasterSource.scala
@@ -33,31 +33,31 @@ case class GDALRasterSource(
 ) extends RasterSource {
   val path: String = dataPath.path
 
-  lazy val datasetType: Int = options.datasetType
+  @transient lazy val datasetType: Int = options.datasetType
 
   // current dataset
   @transient lazy val dataset: GDALDataset =
     GDALDataset(path, options.toWarpOptionsList.toArray)
 
-  lazy val bandCount: Int = dataset.bandCount
+  @transient lazy val bandCount: Int = dataset.bandCount
 
-  lazy val crs: CRS = dataset.crs
+  @transient lazy val crs: CRS = dataset.crs
 
   // noDataValue from the previous step
-  lazy val noDataValue: Option[Double] = dataset.noDataValue(GDALWarp.SOURCE)
+  @transient lazy val noDataValue: Option[Double] = dataset.noDataValue(GDALWarp.SOURCE)
 
-  lazy val dataType: Int = dataset.dataType
+  @transient lazy val dataType: Int = dataset.dataType
 
-  lazy val cellType: CellType = dstCellType.getOrElse(dataset.cellType)
+  @transient lazy val cellType: CellType = dstCellType.getOrElse(dataset.cellType)
 
-  lazy val gridExtent: GridExtent[Long] = dataset.rasterExtent(datasetType).toGridType[Long]
+  @transient lazy val gridExtent: GridExtent[Long] = dataset.rasterExtent(datasetType).toGridType[Long]
 
   /** Resolutions of available overviews in GDAL Dataset
     *
     * These resolutions could represent actual overview as seen in source file
     * or overviews of VRT that was created as result of resample operations.
     */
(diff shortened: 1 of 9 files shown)


gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALRasterSource.scala
```diff
@@ -33,31 +33,31 @@ case class GDALRasterSource(
...
-  lazy val gridExtent: GridExtent[Long] = dataset.rasterExtent(datasetType).toGridType[Long]
+  @transient lazy val gridExtent: GridExtent[Long] = dataset.rasterExtent(datasetType).toGridType[Long]
 
   /** Resolutions of available overviews in GDAL Dataset
     *
     * These resolutions could represent actual overview as seen in source file
     * or overviews of VRT that was created as result of resample operations.
     */
-  lazy val resolutions: List[GridExtent[Long]] = dataset.resolutions(datasetType).map(_.toGridType[Long])
+  @transient lazy val resolutions: List[GridExtent[Long]] = dataset.resolutions(datasetType).map(_.toGridType[Long])
```
pomadchin: For instance, why this is required to be `transient`? it is a serializable value. I would like to see why these values becoming `nulls` after deserialization.

vlm/src/main/scala/geotrellis/contrib/vlm/avro/GeotrellisRasterSource.scala
```diff
@@ -60,19 +60,19 @@ class GeotrellisRasterSource(
     this(AttributeStore(dataPath.path), dataPath)
 
 
-  lazy val reader = CollectionLayerReader(attributeStore, dataPath.path)
+  @transient lazy val reader = CollectionLayerReader(attributeStore, dataPath.path)
 
   // read metadata directly instead of searching sourceLayers to avoid unneeded reads
-  lazy val metadata = reader.attributeStore.readMetadata[TileLayerMetadata[SpatialKey]](layerId)
+  @transient lazy val metadata = reader.attributeStore.readMetadata[TileLayerMetadata[SpatialKey]](layerId)
```
pomadchin: Again the question is why this `metadata` should be `transient`?

gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALRasterSource.scala
pomadchin: Sine `@transient`sets value to `null` during serialization, it is not obvious how it addresses the behavior that is described in #82, since it will cause references to be set to `nulls`. Check out for instance [this source](https://dzone.com/articles/diving-deeper-into-the-java-transient-modifier) to learn more about how `transient` notation works.

pomadchin:
@jbouffard @moradology also tried this small serialization check: 

```scala
      val res =
        sc
          .parallelize(1 to 10, 5)
          .map { _ =>
            val res = GDALRasterSource(uri, GDALWarpOptions(alignTargetPixels = false))
            // res.gridExtent
            res
          }
          .repartition(10)
          .map { rs =>
            (
              rs.gridExtent,
              rs.cellType,
              rs.noDataValue,
              rs.crs,
              rs.noDataValue
            )
          }

      res.collect().toList
```

It works without this PR

jbouffard:
@pomadchin Can this PR just be closed then?

pomadchin:
@jbouffard this PR probably can be closed since it doesn't address the issue; but let's keep it open; it would be good to reproduce the issue @moradology talked about in #82 and to be sure that it is resolved / not our issue.

echeipesh:
Throwing my 2-cents into the hat:

👍 on trying to reproduce the issue with a serialization test. That is the correct thing to do because its extremely error prone to reason about the serialization behavior.

Although not introduced by this PR the `lazy val` usage is a little nuts and jumps out as a sore thumb. As part of solving the underlying problem I would especially investigate how many of them in `GDALRasterSource` can be turned into `def`s since they're delegating to the underlying `dataset`.

jbouffard:
@moradology Isn't able to replicate the issue this PR seeks to solve, so for right now I'm going to close this.

PR #213, opened 2019-07-16
Add Azavea and year to LICENSE

CloudNiner:
Was looking for the correct official name of Azavea for use in LICENSEs, went here and noticed it was missing. So now its updated.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -186,7 +186,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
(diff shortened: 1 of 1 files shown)


pomadchin:
@CloudNiner do we need to update LICENSE [here](https://github.com/locationtech/geotrellis/blob/master/LICENSE)?

CloudNiner:
I think the locationtech/geotrellis license is fine. That calls out the full text in a link and is consistent with other locationtech project apache licenses.

PR #216, opened 2019-07-18
Expose GDAL Metadata API, expose RasterSources Metadata API

### LINKED ISSUE #181: RasterSource Metadata API
RasterSource needs to expose metadata of the underlying rasters through API that does not require side-loading the metadata from format specific readers.

This presents a challenge because current instances of raster source already range over multiple file formats with GDAL and storage formats with GeoTrellis layers. 

Starting expectation is that this Metadata class hierarchy may expose common raster metadata but after that require matching/reflection to determine what is available for specific source. This API should not rely on type parameters since the type information will not be available at runtime for this I/O module.

Some traits that _may_ exist in this hierarchy are:
- `TemporalMetadata`
- `GeoTiffTagsMetadata`
- `RasterMetadata`
        
We should consider if we can re-use some existing API for this purpose:
- OGC `FeatureType`
- OGC `CooordinateReferenceSystem`
- `org.opengis.metadata` package from GeoTools

Above interfaces are Java interfaces and may add too much burden on the user code when using them in Scala (for instance `org.opengis.metadata.MetaData` makes heavy use of java `Collection` which would encourage heavy use of `JavaConverters`. We should still take a look and make a considered choice there.

pomadchin:
# Overview

This PR exposes the metadata that is not covered by the RasterSources API, such as tiff tags / extra metadata that can be interpreted as a string. 

In terms of this PR we've tried to look how other GIS libraries collect imagery metadata, and it looks like there is no a standard interface to work with rather than a `Map[String, String]`.

## Checklist

Closes #181


vlm/src/main/scala/geotrellis/contrib/vlm/RasterSourceMetadata.scala
```diff
@@ -0,0 +1,60 @@
...
+    */
+  def resolutions: List[GridExtent[Long]]
+
+  def extent: Extent = gridExtent.extent
+
+  /** Raster pixel column count */
+  def cols: Long = gridExtent.cols
+
+  /** Raster pixel row count */
+  def rows: Long = gridExtent.rows
```
metasim: FTW! Wrapping all of the above together is fantastic. It allows me to get rid of [this hack](https://github.com/s22s/rasterframes/blob/develop/core/src/main/scala/org/locationtech/rasterframes/ref/SimpleRasterInfo.scala) which was an attempt to cache this info in a serializable format.

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSourceMetadata.scala
```diff
@@ -0,0 +1,60 @@
...
+
+  /** All available metadata that was not covered by other RasterSource metadata methods */
+  def metadata: SourceMetadata
```
metasim: @pomadchin Did you consider `srcMetadata`?

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSourceMetadata.scala
```diff
@@ -0,0 +1,60 @@
...
+ * limitations under the License.
+ */
+
+package geotrellis.contrib.vlm
+
+import geotrellis.proj4.CRS
+import geotrellis.raster.{CellSize, CellType, GridExtent, RasterExtent}
+import geotrellis.vector.Extent
+
+trait RasterSourceMetadata {
```
metasim: FTW! Wrapping all of the below together is fantastic. It allows me to get rid of [this hack](https://github.com/s22s/rasterframes/blob/develop/core/src/main/scala/org/locationtech/rasterframes/ref/SimpleRasterInfo.scala) which was an attempt to cache this info in a serializable format. Can't wait to use it.

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSourceMetadata.scala
```diff
@@ -0,0 +1,60 @@
...
+
+  /** Raster pixel column count */
+  def cols: Long = gridExtent.cols
+
+  /** Raster pixel row count */
+  def rows: Long = gridExtent.rows
+
+  /** All available metadata that was not covered by other RasterSource metadata methods */
+  def metadata: SourceMetadata
+}
```
metasim: If I'm interested in having a serializable (`case` class) implementation of this (not sure what to name it.... `ConcreteRasterSourceMetadata`?), should we add it to this code base, or should the end user be responsible for implementing it?

vlm/src/main/scala/geotrellis/contrib/vlm/SourceMetadata.scala
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package geotrellis.contrib.vlm
+
+trait SourceMetadata {
+  def base: Map[String, String]
+  def band(b: Int): Map[String, String]
+  def combine(self: SourceMetadata, bandCount: Int): SourceMetadata
```
metasim: I'd suggest documenting the semantics of this... not clear to me what the end result would be, what `bandCount` is specifying, what happens if keys overlap etc. IOW, what are the laws :-p

pomadchin: @metasim in addition I also don't like the names as they don't describe semantics; if you have any better suggestions you're welcome!

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSourceMetadata.scala
pomadchin: Nope! Do you think it is better? I also though about it or about `baseMetadata` but sounds for some reason like some sort of tautology

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSourceMetadata.scala
pomadchin: @metasim do you mean that instead of having a `trait` you would like to see just some generic case class that unifies this `metadata` representation?

metasim: @pomadchin Maybe?... kinda a philosophical question.  I guess if the API doesn't need it it shouldn't implement it. But I will need it in RF...

vlm/src/main/scala/geotrellis/contrib/vlm/SourceMetadata.scala
metasim: @pomadchin I'd need to know what it does before suggesting a name!

pomadchin: @metasim I just removed it

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSourceMetadata.scala
pomadchin: @metasim can you show the thing that you need?

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSource.scala
```diff
@@ -42,41 +42,11 @@ import java.util.ServiceLoader
   * @groupdesc reproject Functions to resample raster data in target projection.
   * @groupprio reproject 2
   */
-trait RasterSource extends CellGrid[Long] with Serializable {
+trait RasterSource extends CellGrid[Long] with RasterSourceMetadata with Serializable {
```
pomadchin: Does this mixin makes any sense? `CellGrid[Long] with RasterSourceMetadata`

vlm/src/main/scala/geotrellis/contrib/vlm/RasterSourceMetadata.scala
pomadchin: I've done it; it looks like the metadata class is what you need now

pomadchin:
@echeipesh I am not sure does the commit e4c4db6 makes things better or not :/ need a pair of fresh eyes here.

vlm/src/main/scala/geotrellis/contrib/vlm/SourceMetadata.scala
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package geotrellis.contrib.vlm
+
+trait SourceMetadata extends RasterSourceMetadata {
+  // contains metadata that is not a part of the RasterSourceMetadata
+  def sourceMetadata(): Map[String, String]
```
pomadchin: `()` are here to handle overload.

vlm/src/main/scala/geotrellis/contrib/vlm/avro/GeoTrellisMetadata.scala
```diff
@@ -0,0 +1,47 @@
...
+object GeoTrellisMetadata {
+  def apply[F[_]: Monad](
+    base: F[Map[String, String]],
+    crs: F[CRS],
+    bandCount: F[Int],
+    cellType: F[CellType],
+    gridExtent: F[GridExtent[Long]],
+    resolutions: F[List[GridExtent[Long]]]
+  ): F[GeoTrellisMetadata] =
+    (base, crs, bandCount, cellType, gridExtent, resolutions).mapN(GeoTrellisMetadata.apply)
```
pomadchin: `GeoTrellisMetadata.apply` is explicitly using `apply` method to disambiguate case class and its companion object.

vlm/src/test/scala/geotrellis/contrib/vlm/avro/GeotrellisRasterSourceSpec.scala
```diff
@@ -118,11 +117,10 @@ class GeotrellisRasterSourceSpec extends FunSpec with RasterMatchers with Better
...
       assert(
         sourceMultiband.resolutions.length ===
-          CollectionLayerReader(uriMultibandNoParams).attributeStore.layerIds.filter(_.name == layerId.name).length
-      )
-      assert(
-        new GeotrellisRasterSource(s"$uriMultibandNoParams?layer=bogusLayer&zoom=0").resolutions.length === 0
+          CollectionLayerReader(uriMultibandNoParams).attributeStore.layerIds.count(_.name == layerId.name)
       )
+
+      intercept[AttributeNotFoundError](new GeotrellisRasterSource(s"$uriMultibandNoParams?layer=bogusLayer&zoom=0").resolutions)
```
pomadchin: I want also to admit that this behavior change caused because all the metadata will be fetched at once now.

gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALDataset.scala
```diff
@@ -17,19 +17,95 @@
...
+    getAllMetadata(GDALWarp.SOURCE, band)
+
+  def getAllMetadata(dataset: Int, band: Int): Map[String, Map[String, String]] =
+    getMetadataDomainList(dataset).map { domain => domain -> getMetadata(domain, band) }.filter(_._2.nonEmpty).toMap
+
+  def getMetadataDomainList(band: Int): List[String] = getMetadataDomainList(GDALWarp.SOURCE, band)
+
+  /** https://github.com/OSGeo/gdal/blob/b1c9c12ad373e40b955162b45d704070d4ebf7b0/gdal/doc/source/development/rfc/rfc43_getmetadatadomainlist.rst */
+  def getMetadataDomainList(dataset: Int, band: Int): List[String] = {
+    val arr = Array.ofDim[Byte](100, 1 << 10)
```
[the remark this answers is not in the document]
pomadchin: @metasim I need to benchmark @jamesmcclain bindings; I think it is a completely separate issue now, as probably requires some investigation time

gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALDataset.scala
```diff
@@ -17,19 +17,95 @@
...
+    new String(arr, "UTF-8").trim
+  }
+
   def getProjection: Option[String] = getProjection(GDALWarp.WARPED)
 
   def getProjection(dataset: Int): Option[String] = {
     require(acceptableDatasets contains dataset)
-    val crs = Array.ofDim[Byte](1 << 16)
-
+    val crs = Array.ofDim[Byte](1 << 10)
```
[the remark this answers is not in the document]
pomadchin: Separate issue!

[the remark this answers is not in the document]
pomadchin: https://github.com/geotrellis/geotrellis-contrib/issues/221

gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALDataset.scala
```diff
@@ -17,19 +17,95 @@
...
+      throw new MalformedProjectionException(
+        s"Unable to get the metadata item. GDAL Error Code: $positiveValue",
+        positiveValue
+      )
+    }
+
+    new String(arr, "UTF-8").trim
+  }
+
   def getProjection: Option[String] = getProjection(GDALWarp.WARPED)
```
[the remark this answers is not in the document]
pomadchin: Or mb even opaque types (https://github.com/estatico/scala-newtype); but it is not a part of this PR

gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALMetadata.scala
```diff
@@ -0,0 +1,67 @@
...
+    domains match {
+      case DomainList(domains) if domains.isEmpty =>
+        GDALMetadata(rasterSource.crs, rasterSource.bandCount, rasterSource.cellType, rasterSource.gridExtent, rasterSource.resolutions)
+      case DomainList(domains) =>
+        GDALMetadata(
+          rasterSource.crs, rasterSource.bandCount, rasterSource.cellType, rasterSource.gridExtent, rasterSource.resolutions,
+          dataset.getMetadata(GDALWarp.SOURCE, domains, 0),
+          (1 until dataset.bandCount).toList.map(dataset.getMetadata(GDALWarp.SOURCE, domains, _))
+        )
+      case FullDomainList =>
```
pomadchin: Mb it should be `AllMetadataDomainsList`? Or smth like that.

pomadchin:
Travis is down for some reason :/ in my fork it passes tests https://travis-ci.org/pomadchin/geotrellis-contrib/jobs/566145813

PR #217, opened 2019-07-18
Add a GDALRasterSourcesF implementation

pomadchin:
# Overview

This PR implements GDALRasterSourcesF

## Checklist

### DIFF
--- a/gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALWarpOptions.scala
+++ b/gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALWarpOptions.scala
@@ -24,6 +24,7 @@ import geotrellis.proj4.CRS
 import geotrellis.vector.Extent
 
 import com.azavea.gdal.GDALWarp
+import cats.Monad
 import cats.instances.option._
 import cats.syntax.apply._
 import cats.syntax.option._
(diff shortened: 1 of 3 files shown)


gdal/src/main/scala/geotrellis/contrib/vlm/gdal/effect/GDALRasterSource.scala
```diff
@@ -0,0 +1,152 @@
...
+import com.azavea.gdal.GDALWarp
+
+import cats._
+import cats.syntax.flatMap._
+import cats.syntax.apply._
+import cats.syntax.functor._
+
+case class GDALRasterSource[F[_]: Monad: UnsafeLift](
+  dataPath: GDALDataPath,
+  options: F[GDALWarpOptions] = GDALWarpOptions.EMPTY,
```
pomadchin: @echeipesh it was a bit hard decision to make it like this, but actually we can't get around it :/

gdal/src/main/scala/geotrellis/contrib/vlm/gdal/GDALWarpOptions.scala
```diff
@@ -255,6 +256,8 @@ case class GDALWarpOptions(
 object GDALWarpOptions {
   val EMPTY = GDALWarpOptions()
 
+  implicit def lift2Monad[F[_]: Monad](options: GDALWarpOptions): F[GDALWarpOptions] = Monad[F].pure(options)
```
pomadchin: This implicit actually, makes the fact that options should be wrapped now invisible.

PR #230, opened 2019-08-12
Make baseTiff transient

### LINKED ISSUE #226: RasterSources should be JavaSerializable
Double check that all tests pass under JavaSer; This may require some proj4 fixes (i.e. `BasicCoordinateTransform` is not serializable).

pomadchin:
# Overview

Addresses https://github.com/geotrellis/geotrellis-contrib/issues/226 as well. 
In some cases GeoTiffs would not be serializable due to having non serializable byte buffers. 
Probably a better solution would be to improve GeoTiff class and make it serializable, so it is probably a temporal solution.

Closes #226

### DIFF
--- a/vlm/src/main/scala/geotrellis/contrib/vlm/geotiff/GeoTiffResampleRasterSource.scala
+++ b/vlm/src/main/scala/geotrellis/contrib/vlm/geotiff/GeoTiffResampleRasterSource.scala
@@ -32,13 +32,13 @@ case class GeoTiffResampleRasterSource(
   method: ResampleMethod = NearestNeighbor,
   strategy: OverviewStrategy = AutoHigherResolution,
   private[vlm] val targetCellType: Option[TargetCellType] = None,
-  private[vlm] val baseTiff: Option[MultibandGeoTiff] = None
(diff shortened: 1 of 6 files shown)


vlm/src/main/scala/geotrellis/contrib/vlm/geotiff/GeoTiffResampleRasterSource.scala
```diff
@@ -32,13 +32,13 @@ case class GeoTiffResampleRasterSource(
   method: ResampleMethod = NearestNeighbor,
   strategy: OverviewStrategy = AutoHigherResolution,
   private[vlm] val targetCellType: Option[TargetCellType] = None,
-  private[vlm] val baseTiff: Option[MultibandGeoTiff] = None
+  @transient private[vlm] val baseTiff: Option[MultibandGeoTiff] = None
```
pomadchin: Since `baseTiff` is transient now, to prevent serialization issues; the following case can happen on shuffle: 

```scala
val baseTiff: Option[MultibandGeoTiff] = null
Option(baseTiff).flatten //> None

```

Issue #233, opened 2019-08-19
Upgrade dependencies


jisantuc:
Version 0.11.1 isn't binary compatible with 0.9.1. If a newer project decides to use pureconfig, they'll probably pick the newest version (I just did), and it takes a small but still low-key annoying quantity of effort to figure out where `java.lang.NoSuchMethodError: pureconfig.ConfigReader$.traversableReader(Lpureconfig/Derivation;Lscala/collection/generic/CanBuildFrom;)Lpureconfig/ConfigReader` is coming from

pomadchin:
Let's make it a more generic task: update all deps before the RC1?

jisantuc:
Sensible -- might be easiest to do this by setting up scala steward, which is some work we think we should also do in RF

pomadchin:
Well, in our case it won't help ): we'll have to file CQs ourselves, and this is the problem rather than bumping up deps versions.

PR #234, opened 2019-08-20
Upgrade pureconfig to current

jisantuc:
# Overview

This PR upgrades pureconfig. New projects that depend on gt-contrib are likely to choose new versions of pureconfig. However, pureconfig 0.9.1 and 0.11.1 (and probably some versions in between) are not binary compatible, so that causes a small but still sort of annoying problem.

## Checklist

- [ ] Add entry to CHANGELOG.md

Helps with #233

### DIFF
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -47,7 +47,7 @@ object Dependencies {
 
   val gdal                    = "org.gdal"                    % "gdal"                       % Properties.envOrElse("GDAL_VERSION", "2.4.0")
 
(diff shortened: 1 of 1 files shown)


pomadchin:
@jisantuc I think this is a wrong place for the change, even though it is for some reason in deps, it is not anywhere added: https://github.com/geotrellis/geotrellis-contrib/search?q=pureconfig&unscoped_q=pureconfig

jisantuc:
Ah woops -- I couldn't start sbt in gt-contrib locally due to some error about `Multiple garbage collectors selected` (my machine probably) so I didn't actually get to test.

PR #236, opened 2019-09-25
Publish gt-contrib artifacts automatically to SonaType via CircleCI

### LINKED ISSUE #122: Improve TravisCI builds
Copy the https://github.com/geotrellis/geotrellis-gdal build scripts and logic

### LINKED ISSUE #222: Steamlined release process (like maml and GT Server)
The deployment best practices developed for:

- https://github.com/geotrellis/maml
and
- https://github.com/geotrellis/geotrellis-server/pull/138

have are something we'd like to see for this project as well. Copying the work done there might be sensible, but it would likely be time consuming given the context ops already has.

@echeipesh said @CloudNiner may have started a conversation about this?

But wanted to create this issue to track the conversation.

CloudNiner:
# Overview

Switch from manual Bintray publishing to automated CI-driven publishing to SonaType, using the same general workflow currently in use on https://github.com/geotrellis/geotrellis-server and https://github.com/geotrellis/maml

TODO: 
- [x] Hook up PR UI in GitHub to include a CircleCI "tests passing" check
- [x] Add a docker-compose container for the `benchmark` subproject in order to cleanup running benchmarks
- [ ] ~~Bump to Scala 2.12.10?~~

## Demo

Screen Shot 2019-09-25 at 1 48 35 PM

## Notes

See discussion on original issue regarding the choice of the `machine` executor that just runs the new local docker-compose setup: https://github.com/geotrellis/geotrellis-contrib/issues/222#issuecomment-535041064

As part of this effort, I updated this repo to default to `develop` and protected both the `develop` and `master` branch. 

I removed the TravisCI webhooks for this repo from the repo settings in GitHub. I also turned off all toggles that I could get to from the TravisCI dashboard. I believe TravisCI is purged from this repo, but it would be worth double-checking.

Kudos to @rbreslow for constructing a straightforward, reusable workflow to follow for this effort.

The first publish task I attempted ran into a 503 Service Temporarily Unavailable error for sonatype. Not sure how transient those are...  https://circleci.com/gh/geotrellis/geotrellis-contrib/17

## Testing Instructions

`./scripts/test` should complete, running vlm, gdal and slick tests inside Docker via docker-compose. 

`./scripts/benchmark` should build and begin running. Once its running, feel free to kill, as on a typical laptop processor it will take a very long time (> 1hr) to complete.

The build for this branch should be green: https://circleci.com/gh/geotrellis/geotrellis-contrib/tree/feature%2Fawf%2Fci

Closes #222, Closes #122


CloudNiner:
@geotrellis/operations this is ready for an initial look. All the core work is done AFAIK.

build.sbt
```diff
@@ -36,32 +42,61 @@ lazy val commonSettings = Seq(
...
   ),
-  headerLicense := Some(HeaderLicense.ALv2("2019", "Azavea")),
-  bintrayOrganization := Some("azavea"),
-  bintrayRepository := "geotrellis",
-  bintrayPackageLabels := Seq("gis", "raster", "vector"),
-  bintrayReleaseOnPublish := false,
-  publishTo := {
-    val bintrayPublishTo = publishTo.value
-    val nexus = "http://nexus.internal.azavea.com"
+  headerLicense := Some(HeaderLicense.ALv2("2019", "Azavea"))
```
rbreslow: I would consider moving this under the `publishSettings` block, as is done in [geotrellis-gdal](https://github.com/geotrellis/geotrellis-gdal/blob/a78055c75522d7c14ebde9c0cdc371ed7b837e6a/build.sbt#L53), or consider removing sbt-header altogether. 

It [doesn't look like](https://github.com/geotrellis/geotrellis-server/search?q=sbt-header&type=Issues) it is used in geotrellis-server or [in](https://github.com/geotrellis/maml/search?utf8=✓&q=headerLicense&type=) maml.

docker-compose.yml
```diff
@@ -0,0 +1,47 @@
...
+  database:
+    # When this is updated, we'll have to handle creating the postgis extension
+    # somewhere in the test workflow. That's handled automatically on 0.1.0 but
+    # not on later versions
+    image: quay.io/azavea/postgis:0.1.0
+    environment:
+      - POSTGRES_USER=slick_tests
+      - POSTGRES_PASSWORD=slick_tests
+      - POSTGRES_DB=slick_tests
+    healthcheck:
```
rbreslow, line 12: Not sure if this `healthcheck` block is particularly useful since Docker dropped support for depending on another service to become healthy in file format 3+ (https://github.com/docker/compose/issues/4305).

A potential workaround would be to pin to version 2.3 so we could do:

```yml
depends_on:
  database:
    condition: service_healthy
```

.circleci/config.yml
```diff
@@ -0,0 +1,104 @@
...
+          gpg --batch \
+            --passphrase "${GPG_PASSPHRASE}" \
+            --import signing_key.asc
+    - run:
+        name: Executing cipublish
+        command: ./scripts/cipublish
+
+  # Build environments
+  - &openjdk8-scala2_11_12_environment
+    machine:
```
rbreslow, line 38: I read through your evaluation on how to approach utilizing the PostGIS database for tests on CircleCI. I think the direction you took is reasonable, and seems to work well. Nice job.

scripts/benchmark
```diff
@@ -0,0 +1,26 @@
...
+
+if [[ "${BASH_SOURCE[0]}" == "${0}" ]]; then
+    if [[ "${1:-}" == "--help" ]]; then
+        usage
+    else
+        echo "Executing GeoTrellis Contrib benchmarks"
+        pushd benchmark
+        make
+        popd
+        docker-compose run --rm benchmark "project benchmark" "jmh:run"
```
rbreslow: Nitpick-y. I'd consider formatting like this when invoking `docker-compose`:

```bash
docker-compose \
  run --rm benchmark \
  "project benchmark" "jmh:run"

rbreslow (approved):
I'm glad the workflow was pretty easy to decipher and carry over from maml and geotrellis-server.

Tests are passing. `./scripts/benchmark` is able to produce a container image and begin running benchmarks.

A few minor comments, but otherwise, looks good 👍

.circleci/config.yml
CloudNiner: Thanks! We'll see how it holds up.

build.sbt
CloudNiner: Good call, I'll move it.

docker-compose.yml
CloudNiner: Shoot. I left myself a TODO and then missed this. The compose docs [have a page on dealing with this](https://docs.docker.com/compose/startup-order/) but if we used one of those wrapper scripts, I _think_ the most straightforward way to use it would be in the `scripts/` or to modify the container entrypoints. But, there's likely some case where you don't want it and its extra code to manage. 

So, I think I'll just revert to compose 2.3 for now. They haven't dropped support for an old version of the compose file format since 1.0 yet.

PR #238, opened 2019-09-27
Fix CI publishing failing with Java SSL handshake exceptions to repo.locationtech.org

CloudNiner:
## Overview

`./scripts/cipublish` uses native java + ./sbt because otherwise we need to install and mount gpg information in the sbt container when publishing, which is a hassle. Native Java on the latest machine executor available in CircleCI happens to have a version of Java too old to have the cipher repo.locationtech.org uses. So here we update the version of Java on the CI machine executor rather than do the configuration necessary to wrap GPG and our keys up in a container with SBT. Code comments and git log have an extra bit of detail.

Connects #222

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,15 +34,26 @@ aliases:
         command: ./scripts/cipublish
 
   # Build environments
-  - &openjdk8-scala2_11_12_environment
+  - &machine-openjdk8-scala2_11_12-environment
     machine:
       image: ubuntu-1604:201903-01
     environment:
       SCALA_VERSION: 2.11.12
 
-  - &openjdk8-scala2_12_8_environment
+  - &machine-openjdk8-scala2_12_8-environment
     machine:
       image: ubuntu-1604:201903-01
+    environment:
+      SCALA_VERSION: 2.12.8
+  - &openjdk8-scala2_11_12-environment
+    docker:
+      - image: circleci/openjdk:8-jdk
+    environment:
+      SCALA_VERSION: 2.11.12
+
+  - &openjdk8-scala2_12_8-environment
+    docker:
+      - image: circleci/openjdk:8-jdk
     environment:
       SCALA_VERSION: 2.12.8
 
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -73,6 +91,7 @@ workflows:
                 - develop
                 - /release\/.*/
                 - /hotfix\/.*/
+                - feature/awf/fix-publish
```
CloudNiner: TODO: Remove before merging

.circleci/config.yml
```diff
@@ -85,6 +104,7 @@ workflows:
...
```
CloudNiner: TODO: Remove before merging

.circleci/config.yml
```diff
@@ -20,6 +20,26 @@ aliases:
 
   - &run_cipublish
     - checkout
+    - run:
```
rbreslow: If we take this approach, I would consider breaking these steps out into their own anchor. Something like `&update_machine_java`.

rbreslow (commented):
What are your thoughts on executing `cipublish` with a `docker` executor? It looks like we only need to utilize `docker-compose` in the build job. The cache would still be restored, since it should be indifferent to the executor type.

Here's a patch:

```diff
diff --git a/.circleci/config.yml b/.circleci/config.yml
index a0a4b84..cfd09e8 100644
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -54,17 +54,28 @@ aliases:
         command: ./scripts/cipublish

   # Build environments
-  - &openjdk8-scala2_11_12_environment
+  - &machine-openjdk8-scala2_11_12_environment
     machine:
       image: ubuntu-1604:201903-01
     environment:
       SCALA_VERSION: 2.11.12

-  - &openjdk8-scala2_12_8_environment
+  - &machine-openjdk8-scala2_12_8_environment
     machine:
       image: ubuntu-1604:201903-01
     environment:
       SCALA_VERSION: 2.12.8
+  - &openjdk8-scala2_11_12-nodelts_environment
+    docker:
+      - image: circleci/openjdk:8-jdk
+    environment:
+      SCALA_VERSION: 2.11.12
+
+  - &openjdk8-scala2_12_8-nodelts_environment
+    docker:
+      - image: circleci/openjdk:8-jdk
+    environment:
+      SCALA_VERSION: 2.12.8

 version: 2
 workflows:
@@ -110,11 +121,11 @@ workflows:

 jobs:
   "openjdk8-scala2.11.12":
-    <<: *openjdk8-scala2_11_12_environment
+    <<: *machine-openjdk8-scala2_11_12_environment
     steps: *run_cibuild

   "openjdk8-scala2.12.8":
-    <<: *openjdk8-scala2_12_8_environment
+    <<: *machine-openjdk8-scala2_12_8_environment
     steps: *run_cibuild

   "openjdk8-scala2.11.12_deploy":
```

CloudNiner:
@rbreslow just applied your patch and pushed a new commit, we'll see if it works. I'd expect it to since its the same workflow as the other repos and the docker image has a newer version of docker installed.

> What are your thoughts on executing cipublish with a docker executor?

Overall, I prefer this solution (assuming it works). It brings this setup more in line with the others, and reduces our dependency on the machine images which are slower.

rbreslow (approved):
👍

